PR #19, opened 2018-07-04
Added new feed layout with comments

hcarneiro:


vendor/autosize.js
```diff
@@ -0,0 +1,295 @@
+/*!
```
squallstar: is it worth having this as a api package?

hcarneiro: @squallstar Wouldn't hurt I guess. Only chat and now this use it though

PR #46, opened 2018-09-05
Adds "None" option when columns are refreshed

hcarneiro:
Ref: https://github.com/Fliplet/fliplet-studio/issues/3139

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -951,6 +951,7 @@ var DynamicLists = (function() {
         $(obj).html('');
         $(obj).append('<option value="none">-- Select a data field</option>');
         $(obj).append('<option disabled>------</option>');
+        $(obj).append('<option value="empty">None</option>');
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -951,6 +951,7 @@ var DynamicLists = (function() {
         $(obj).html('');
         $(obj).append('<option value="none">-- Select a data field</option>');
         $(obj).append('<option disabled>------</option>');
+        $(obj).append('<option value="empty">None</option>');
         $(obj).append('<option value="custom">Custom</option>');
         $(obj).append('<option disabled>------</option>');
```
tonytlwu: @hcarneiro The `$(obj)` element is being updated 7 times (`.html('')` x 1, `.append()` x 5, and then another `.append()` at the end. You should aim to do this only once for each `.each()` loop.

PR #53, opened 2018-09-20
Make sure we can always split via a comma

hcarneiro:
Ref: https://github.com/Fliplet/fliplet-studio/issues/3242


tonytlwu:
@hcarneiro Since it's a split-by-comma operation, have you considered using https://github.com/Fliplet/fliplet-widget-data-directory/blob/master/js/build-directory.js#L18-L36 ?

This allows commas to be excluded if the **term/tag** is encapsulated in `"`.

e.g. `a tag, another tag, "a tag with a , inside", one more tag`

PR #54, opened 2018-09-20
Removes the automatic data source creation

hcarneiro:
This removes the automatic creation of data source triggered after selecting a list style.
Instead it loads the based on the data that we used before to create the data source.

Now the user has to create a new data source, instead of seeing the "Edit" and "Change" data source right away.

screen shot 2018-09-20 at 18 24 29


templates/build/agenda-cards-loop.build.hbs
```diff
@@ -19,7 +19,7 @@
                   \{{#if [Location]}}
                   <p class="agenda-item-location">\{{{[Location]}}}</p>
                   \{{/if}}
-                  <p class="agenda-item-start-date-time">\{{[Start Time]}}\{{#ifCond [Start Time] '&&' [End Time]}} - \{{/ifCond}}\{{[End Time]}}\{{#if [Date]}} / \{{/if}}\{{[Date]}}</p>
+                  <p class="agenda-item-start-date-time">\{{[Start Time]}}\{{#ifCond [Start Time] '&&' [End Time]}} - \{{/ifCond}}\{{[End Time]}}\{{#if [Full Date]}} / \{{/if}}\{{formatDate [Full Date]}}</p>
```
tonytlwu: Hopefully you won't need this soon because we're working on the `moment` helper. That would help, right?

hcarneiro: @tonytlwu I won't need that helper, I'll update to the one you create(d).

PR #61, opened 2018-10-10
Review comments

tonytlwu:
Reviewed after an issue almost prevented a Freshfields app from launching.

Please make sure the other LFD layout JS code are reviewed in the same way as well.


hcarneiro:
ref: https://github.com/Fliplet/fliplet-studio/issues/3330

PR #62, opened 2018-10-15
Added ability to limit the number of entries to render

hcarneiro:
For screenshots and more info: https://github.com/Fliplet/fliplet-studio/issues/3287


js/interface.js
```diff
@@ -657,6 +661,7 @@ var DynamicLists = (function() {
               case 'simple-list':
                 $('.filter-loop-item').removeClass('hidden');
                 $('.detail-view-item').removeClass('hidden');
+                $('.items-number').removeClass('hidden');
```
tonytlwu:
```js
            switch(listLayout) {
              case 'small-card':
              case 'news-feed':
              case 'simple-list':
                $('.filter-loop-item, .detail-view-item, .items-number').removeClass('hidden');
                break;
              case 'agenda':
                $('.date-loop-item, .detail-view-item').removeClass('hidden');
                break;
              case 'small-h-card':
                $('.detail-view-item').removeClass('hidden');
                break;
              default:
                break;
            }
```

PR #66, opened 2018-10-30
Solve concurrency issue when resolving link providers

hcarneiro:
Reference of issue ticket: https://github.com/Fliplet/fliplet-e2e-tests/issues/176

### DIFF
--- a/js/interface-lists.js
+++ b/js/interface-lists.js
@@ -224,10 +224,14 @@ function attahObservers() {
           }
         }
 
-        return Promise.all([
-          linkAddEntryProvider.forwardSaveRequest(),
-          linkEditEntryProvider.forwardSaveRequest()
-        ]);
+        var addEntryLinkPromise = new Promise(function(resolve, reject) {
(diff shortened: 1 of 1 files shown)


js/interface-lists.js
```diff
@@ -224,10 +224,14 @@ function attahObservers() {
           }
         }
 
-        return Promise.all([
```
squallstar: ehm... why not simply

```
return linkAddEntryProvider.forwardSaveRequest().then(function () {
  return linkEditEntryProvider.forwardSaveRequest();
});
```

js/interface-lists.js
```diff
@@ -224,10 +224,14 @@ function attahObservers() {
...
-          linkAddEntryProvider.forwardSaveRequest(),
```
squallstar: `forwardSaveRequest` does not return a promise https://github.com/Fliplet/fliplet-api/blob/afc71fa4edfbac0e8ff568141a8a4dd6a34573dc/public/assets/fliplet-core/1.0/core.js#L1173-L1181

PR #77, opened 2018-11-15
Adds ID to settings UI and exposes ID in hooks

hcarneiro:
This adds an ID to the settings UI and exposes the component ID on the hooks.
screenshot 2018-11-15 at 11 54 34

Enhancement made to make easy for CS to train clients on how to use Hooks to target specific LFDs on a page.

### DIFF
--- a/css/interface.css
+++ b/css/interface.css
@@ -9,9 +9,24 @@
   background-color: transparent;
 }
 
+header {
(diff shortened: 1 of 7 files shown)


hcarneiro:
@nvalbusa thanks I'll do that then.

PR #85, opened 2018-12-10
Added optimizations to layouts.

angelnikolov:
tl;dr https://www.useloom.com/share/cf12d0da214a4237830884b4bbbef5d8

Based on the discussion and ideas shared in https://github.com/Fliplet/fliplet-studio/issues/3582 I added some optimisations to the LFD so now it should load faster.

The largest performance hit we took was basically in a couple of quadratic complexity nested loops which were unneeded.
The next step to optimise was to reduce 2 unneeded sort/orderBy statements to one.
The final step was to introduce incremental rendering via [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame). In short, we batch all the records by 100s (configurable at code level) and render those batches to the screen whenever the browser gives us the opportunity to do so.
What that means is that we leave the browser some time to render the initial records, then when it's idle it can render the next ones and so on. 

**I will add some review comments on the places which needed optimisations**
### Development experience notes
Time to figure out where the perf hit was about 20 minutes with the use of Chrome's performance audit tools.
The time it took me to figure out where in code that was coming from and more importantly - why - took me about an hour and a half, due to the bloated and unseparated code.
Then, another 3 hours were spent to actually applying the fix to all components (which were all the same, but also somehow different) and test them out.

What I am saying is that all this would've been probably done in less than an hour if we had a consistent framework which has separation of concerns to its core.

### DIFF
--- a/js/layout-javascript/agenda-code.js
+++ b/js/layout-javascript/agenda-code.js
@@ -53,6 +53,12 @@ var DynamicList = function(id, data, container) {
   this.pvPreFilterQuery;
   this.pvOpenQuery;
 
+  /**
+   * this specifies the batch size to be used when rendering in chunks
+   */
+  this.INCREMENTAL_RENDERING_BATCH_SIZE = 100;
+
+
   this.detailHTML = this.data.advancedSettings && this.data.advancedSettings.detailHTML
   ? Handlebars.compile(this.data.advancedSettings.detailHTML)
   : Handlebars.compile(Fliplet.Widget.Templates[_this.agendaLayoutMapping[this.data.layout]['detail']]());
@@ -536,15 +542,13 @@ DynamicList.prototype.filterRecords = function(records, filters) {
 
 DynamicList.prototype.prepareData = function(records) {
   var _this = this;
-  var sorted;
-  var ordered;
   var filtered;
 
   // Prepare sorting
   if (_this.data.sortOptions.length) {
     var fields = [];
     var sortOrder = [];
-    var columns = [];
+    var sortColumns = [];
 
     _this.data.sortOptions.forEach(function(option) {
       fields.push({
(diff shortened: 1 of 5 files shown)


js/layout-javascript/agenda-code.js
```diff
@@ -587,17 +591,17 @@ DynamicList.prototype.prepareData = function(records) {
...
         }
 
-        columns.push('data[modified_' + field.column + ']');
       });
 
       return record;
     });
 
-    // Sort data
-    sorted = _.sortBy(mappedRecords, columns);
```
angelnikolov: Unneeded sort was taking about ~2.5 seconds for 2000 records

js/layout-javascript/agenda-code.js
```diff
@@ -587,17 +591,17 @@ DynamicList.prototype.prepareData = function(records) {
           record.data['modified_' + field.column] = record.data['modified_' + field.column];
         }
 
-        columns.push('data[modified_' + field.column + ']');
```
angelnikolov: For a 3-field pre-sorted instance, this grew to about ~6700 instances of the same 3 column names, which later resulted in exponentially more loops in the `orderBy`

js/layout-javascript/agenda-code.js
```diff
@@ -902,51 +906,46 @@ DynamicList.prototype.prepareToRenderLoop = function(rows) {
...
     });
 
-    loopData.push(newObject);
-  });
-
-  // Define detail view data based on user's settings
-  var detailData = [];
-
-  dynamicData.forEach(function(obj) {
-    clonedRecords.some(function(entryData) {
```
angelnikolov: We already have access to the clonedRecords iteratee in the loop above, no need to re-do it here and even nested in another loop. Another second shaved off.

js/layout-javascript/agenda-code.js
```diff
@@ -982,14 +981,36 @@ DynamicList.prototype.prepareToRenderLoop = function(rows) {
...
 
-  _this.$container.find('#agenda-cards-wrapper-' + _this.data.id + ' .agenda-list-holder').html(template(_this.modifiedRecords));
+  if (_this.data.enabledLimitEntries && _this.data.limitEntries >= 0 && !_this.isSearching && !_this.isFiltering) {
+    limitedList = _this.modifiedListItems.slice(0, _this.data.limitEntries);
+  }
+  _this.$container.find('#agenda-cards-wrapper-' + _this.data.id + ' .agenda-list-holder').empty();
+
+  var renderLoopIndex = 0;
+  function render() {
+    // get the next batch of items to render
```
angelnikolov: This is the most rudimentary async rendering approach. I still think that we can further optimise this by introducing a real virtual scrolling.

tonytlwu: @angelnikolov Did not know this trick. Awesome!

PR #86, opened 2018-12-18
Optimization fixes

angelnikolov:
Just a proposal for some fixes around the LFD optimizations.
My initial perf improvements involved partial rendering, but I didn't take into account that if we render partially, we have to apply event handlers and transformations, also incrementally.
I introduced two new callbacks to the async renderer which will be called upon every render iteration and once - after the rendering is complete. 
**This is not ready, just putting it out for a discussion**


templates/build/news-feed-base.build.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="new-news-feed-list-container loading">
+<div class="new-news-feed-list-container ready">
```
hcarneiro: You removed the initial loading state? Why?

hcarneiro:
@angelnikolov Please also take into account, that when filtering and searching we re-render the list, and some elements may need to be destroyed and they will need to be re-initialised. Like bookmarks, comments, likes...

There's also a `overrideSearchData()` that needs to be taken into account, as that search happens programatically using the PV query vars.

templates/build/news-feed-base.build.hbs
angelnikolov: That was a test, gonna get it back and just remove the loading state during the first iteration only. Nice catch!

angelnikolov:
@hcarneiro I've tested all filterable layouts the way you showed me and they work.

js/layout-javascript/agenda-code.js
```diff
@@ -1247,10 +1246,10 @@ DynamicList.prototype.prepareToRenderLoop = function(rows) {
     return row[dateField];
   }));
 
-  _this.modifiedRecords = newRecords;
+  _this.agendasByDay = newRecords;
 }
 
-DynamicList.prototype.renderLoopHTML = function(records) {
+DynamicList.prototype.renderLoopHTML = function(iterateeCb, finishCb) {
```
tonytlwu: @angelnikolov Why not use a Promise? We haven't been using callbacks for a while. Most of our code has moved on to use Promises.

angelnikolov: I didn't want to mix up between iterative callback and a promise so the API is more obvious. I can easily change that to a Promise if everyone agrees but I think it's cleaner this way.

tonytlwu: I see.

I think it's best to use a Promise instead of the `finishCb`. The `iteratee` can be kept as a parameter. This way when `.renderLoopHTML()` can use `.then()` and `.catch()` for errors and so on. We've been relying on that for many things.

It'll be odder to change to a `finishCb` parameter half way through a bunch of `.then()` chains.

angelnikolov: Okay, will change those to promises.

angelnikolov: Done, @tonytlwu would you mind checking it again?:) Thanks!

PR #90, opened 2018-12-19
Fixes and improvements for comments

hcarneiro:
- Fixes UI bugs when switching on and off the permissions' settings
- New fields to select folder for comments' user's photo
- Get the file from folder on the two styles that have comments

### DIFF
--- a/js/layout-javascript/simple-list-code.js
+++ b/js/layout-javascript/simple-list-code.js
@@ -949,65 +949,9 @@ DynamicList.prototype.prepareData = function(records) {
   return records;
 }
 
-DynamicList.prototype.convertFilesForComments = function(users) {
+DynamicList.prototype.convertFiles = function(listItems, forComments) {
   var _this = this;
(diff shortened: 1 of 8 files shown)


js/layout-javascript/simple-list-code.js
```diff
@@ -949,6 +949,63 @@ DynamicList.prototype.prepareData = function(records) {
   return records;
 }
 
+DynamicList.prototype.convertFilesForComments = function(users) {
```
angelnikolov: Maybe our next task about optimizing LFD will be more of a development optimization one and it could be to split all the common/repeated functions out in a separate module and import it through webpack @hcarneiro  @squallstar @tonytlwu

hcarneiro: I could probably improve that specific function you highlighted to use a similar that already exists. Just probably need a new parameter to make the distinction. Let me see what I can do.

PR #91, opened 2018-12-20
Changed way of configuring the Poll, Survey, Questions buttons on the Agenda

hcarneiro:
User selects the column, and in the column writes the screen name to where it should be redirected.
_(As per requirements given by CS)_

screenshot 2018-12-20 at 15 53 42

### DIFF
--- a/js/layout-javascript/agenda-code.js
+++ b/js/layout-javascript/agenda-code.js
@@ -53,6 +53,8 @@ var DynamicList = function(id, data, container) {
   this.pvPreFilterQuery;
   this.pvOpenQuery;
 
+  this.appPages;
+
   /*
    * this specifies the batch size to be used when rendering in chunks
    */
@@ -67,18 +69,23 @@ var DynamicList = function(id, data, container) {
   // Register handlebars helpers
   this.registerHandlebarsHelpers();
   // Get the current session data
-  Fliplet.Session.get().then(function(session) {
-    if (session && session.entries && session.entries.dataSource) {
-      _this.myUserData = session.entries.dataSource.data;
-    } else if (session && session.entries && session.entries.saml2) {
-      _this.myUserData = session.entries.saml2.user;
-      _this.myUserData[_this.data.userEmailColumn] = _this.myUserData.email;
-      _this.myUserData.isSaml2 = true;
-    }
-    
-    // Start running the Public functions
-    _this.initialize();
-  });
+  Fliplet.Session.get()
+    .then(function(session) {
+      if (session && session.entries && session.entries.dataSource) {
+        _this.myUserData = session.entries.dataSource.data;
+      } else if (session && session.entries && session.entries.saml2) {
+        _this.myUserData = session.entries.saml2.user;
+        _this.myUserData[_this.data.userEmailColumn] = _this.myUserData.email;
+        _this.myUserData.isSaml2 = true;
+      }
+
+      return Fliplet.Pages.get();
+    })
+    .then(function (appPages) {
+      _this.appPages = appPages;
+      // Start running the Public functions
+      _this.initialize();
+    });
 };
 
 DynamicList.prototype.registerHandlebarsHelpers = function() {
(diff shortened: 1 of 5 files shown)


js/layout-javascript/agenda-code.js
```diff
@@ -156,56 +159,80 @@ DynamicList.prototype.attachObservers = function() {
...
         id: entryId,
         title: entryTitle
       }
+      var entry = _.find(_this.listItems, function(entry) {
+        return entry.id === entryId;
+      });
+      var screenFromColumn = entry.data[_this.data.pollColumn];
+      var screen = _.find(_this.appPages, function(page) {
+        return page.title === screenFromColumn;
+      });
```
tonytlwu: @squallstar This chunk of code is basically going through app pages to find a page ID. I've seen this way too many times. Perhaps we can enhance `Fliplet.Navigate.screen()` to take on the first parameter either as a page ID (number) or a page title (string). Though, the page title (string) might be worth parsing into an integer if there's no page title match, in case a page ID is passed as a string.

What do you think?

tonytlwu: It also means that some of our custom code and app templates will be easier to manage. Right now we sometimes use hardcoded `pageId`s. If they were hardcoded page titles, then those references won't need updating ☺️

hcarneiro: @squallstar @tonytlwu I'm also worried that by using directly in the code `screen.id` might not work on production apps. Because I noticed that each page also has the `masterPageId ` key.

Is this an issue?

squallstar: @tonytlwu I've always been reluctant on having the title matching in our core APIs simply because it is prone to break when the user changes the screen's name.

The code I'd be okay to add is a helper function like `Fliplet.Pages.toID('Foo')` which attempt to convert the title to its ID. What do you think?

squallstar: @hcarneiro no, that wouldn't be an issue. You're more than welcome to use any of the IDs of a page as the mapping to the correct ID is done internally by the JS API as per [this piece of code](https://github.com/Fliplet/fliplet-api/blob/master/public/assets/fliplet-core/1.0/navigate.js#L655-L669)

tonytlwu: @squallstar I understand the concern. I think if `Fliplet.Navigate.screen()` suitably rejects the Promise with an error `Page "${title}" not found` then it would be sufficient. I see it as the same category of errors users would make if they use the wrong page ID.

You can still create a `Fliplet.Pages.toID()` helper if you want, but I'd prefer if `Fliplet.Navigate.screen()` supports page title without needing to use `Fliplet.Navigate.screen(Fliplet.Pages.toID())` 🙈

squallstar: @tonytlwu what do you do when a page has title "2" and ID 1, and `Fliplet.Navigate.screen("1")` is called?

tonytlwu: @squallstar 

> Though, the page title (string) might be worth parsing into an integer if there's no page title match, in case a page ID is passed as a string.

I think it could follow a flow like below so that if a string is passed it always considers it as a title first, and parses it into a number as a fallback.

```js
if (typeof pageId === 'string') {
  var title = pageId;
  pageId = Fliplet.Pages.toID(title)
  if (!pageId) {
    pageId = parseInt(title, 10);
  }
}
```

squallstar: @tonytlwu it sounds like a breaking change to me, because I'm fairly confident there's occurrences in our system (apps, custom code, widgets and so) which sometimes pass the ID as a string.

If we were to support the title as input, parsing the string as a number SHOULD come first to avoid breaking changes to existing code across the platform.

However, as I pointed out in my previous comment that brings inconsistencies when you name a page with a number and want to refer to it (given the same ID is being used by another page). It's a pretty uncommon scenario but worth to be mentioned.

tonytlwu: @squallstar 

> [...] fairly confident there's occurrences in our system (apps, custom code, widgets and so) which sometimes pass the ID as a string

Good point. In that case, I think we have no choice but to use `Fliplet.Pages.toID()` 🤷‍♂️

squallstar: @tonytlwu the alternative is:

- update `.screen()` to search for a page by title given the input is a string and such string (converted to number) was not found in the appPages (or input isNaN)
- if the user has a page title with just a number in it, it can use the `toID()` helper when calling the `.screen()` function to make sure the title gets converted to an ID (and avoid the issues we just talked about)

tonytlwu: @squallstar I see. Yes, I think that sounds like the best of both worlds ☺️

tonytlwu: @squallstar See https://github.com/Fliplet/fliplet-studio/issues/3231

PR #92, opened 2018-12-21
Added bookmarks to directory layout style

hcarneiro:
screenshot 2018-12-21 at 17 08 50
screenshot 2018-12-21 at 17 09 03


js/layout-javascript/small-card-code.js
```diff
@@ -732,6 +738,16 @@ DynamicList.prototype.prepareData = function(records) {
...
   }
 
+  // Add flag for likes
+  records.forEach(function(obj, i) {
+    // Add bookmarks flag
+    if (_this.data.social && _this.data.social.bookmark) {
+      records[i].bookmarksEnabled = true;
+    } else {
+      records[i].bookmarksEnabled = false;
+    }
```
tonytlwu: `records[i].bookmarksEnabled = _this.data.social && _this.data.social.bookmark`

...should be able to replace these 4 lines

js/layout-javascript/small-card-code.js
```diff
@@ -262,6 +268,11 @@ DynamicList.prototype.attachObservers = function() {
     })
     .on('click', '.small-card-list-item', function(event) {
       var _that = $(this);
+
+      if ($(event.target).hasClass('small-card-bookmark-holder') || $(event.target).parents('.small-card-bookmark-holder').length) {
```
tonytlwu: I think in this case `$(this)` is the same as `$(event.target)`, which means you already have the jQuery object cached. Just use `_that.hasClass()` and `_that.parents()`.

PR #99, opened 2019-01-14
Get master back to up to date code

hcarneiro:
This PR includes a bunch of other PR's that were approved before but we had to revert to a certain point in time.

This includes the following features:
- Allow users to select a folder for any field that can be displayed as an Image
- Load optimisations from @angelnikolov 
- Added bookmarks to directory style
- Added a new Agenda feature to add a Poll, Survey, Questions buttons


tonytlwu:
@hcarneiro How about the hooks? Are all the ones we added present in this latest PR/version?

tonytlwu:
Yup, I see some of 'em here. Annoying that they're hidden in **Load diff**, but they seem to be there.

PR #101, opened 2019-01-16
Only get columns if detailViewAutoUpdate is true and device is online

hcarneiro:
This will fix an issue that would prevent the list from rendering when the device is offline.

### DIFF
--- a/js/layout-javascript/agenda-code.js
+++ b/js/layout-javascript/agenda-code.js
@@ -706,10 +706,16 @@ DynamicList.prototype.initialize = function() {
       return;
     })
     .then(function() {
-      return Fliplet.DataSources.getById(_this.data.dataSourceId);
+      if (_this.data.detailViewAutoUpdate && Fliplet.Navigator.isOnline()) {
+        return Fliplet.DataSources.getById(_this.data.dataSourceId);
+      }
+      return
     })
     .then(function(dataSource) {
-      _this.dataSourceColumns = dataSource.columns;
+      if (dataSource) {
+        _this.dataSourceColumns = dataSource.columns;
+      }
+
       return
     })
     .then(function() {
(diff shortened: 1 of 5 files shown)


js/layout-javascript/agenda-code.js
```diff
@@ -706,10 +706,16 @@ DynamicList.prototype.initialize = function() {
       return;
     })
     .then(function() {
-      return Fliplet.DataSources.getById(_this.data.dataSourceId);
+      if (_this.data.detailViewAutoUpdate && Fliplet.Navigator.isOnline()) {
+        return Fliplet.DataSources.getById(_this.data.dataSourceId);
+      }
+      return
```
squallstar: this seems unnecessary?

js/layout-javascript/agenda-code.js
```diff
@@ -706,10 +706,16 @@ DynamicList.prototype.initialize = function() {
       return;
     })
     .then(function() {
-      return Fliplet.DataSources.getById(_this.data.dataSourceId);
+      if (_this.data.detailViewAutoUpdate && Fliplet.Navigator.isOnline()) {
+        return Fliplet.DataSources.getById(_this.data.dataSourceId);
```
squallstar: I would suggest a `.catch(function () { return Promise.resolve(); });`

hcarneiro: But if the error comes from somewhere else it will resolve too.

angelnikolov: I thin he meant adding that to the `getById` call. Right?

squallstar: @hcarneiro I'm suggesting catching this promise, not others.

```js
return Fliplet.DataSources.getById(_this.data.dataSourceId).catch(function () {
  return Promise.resolve(); // no problem when this fails
});
```

PR #110, opened 2019-01-30
LFD fixes

hcarneiro:
- Fixes a UX issue when editing a comment
- Allows Agenda layout to initialise even if no entries are found - https://github.com/Fliplet/fliplet-studio/issues/3763

### DIFF
--- a/js/layout-javascript/agenda-code.js
+++ b/js/layout-javascript/agenda-code.js
@@ -1346,52 +1346,55 @@ DynamicList.prototype.renderDatesHTML = function(rows, index) {
   if (foundDateField) {
     dateField = foundDateField.column;
   }
-  // set first date in agenda
-  firstDate = new Date(clonedRecords[0].data[dateField]).toUTCString();
 
-  // set last date in agenda
-  lastDate = new Date(clonedRecords[clonedRecords.length - 1].data[dateField]).toUTCString();
+  if (clonedRecords.length) {
+    // set first date in agenda
+    firstDate = new Date(clonedRecords[0].data[dateField]).toUTCString();
 
-  // Adds 5 days before the first date
-  // Save them in an array
-  for (var i = 0; i < numberOfPlacholderDays; i++) { 
-    var newDate = {
-      week: moment(firstDate).utc().subtract(i + 1, 'days').format("ddd"),
-      day: moment(firstDate).utc().subtract(i + 1, 'days').format("DD"),
-      month: moment(firstDate).utc().subtract(i + 1, 'days').format("MMM"),
-      placeholder: true
+    // set last date in agenda
+    lastDate = new Date(clonedRecords[clonedRecords.length - 1].data[dateField]).toUTCString();
+
+    // Adds 5 days before the first date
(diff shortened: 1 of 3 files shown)


js/layout-javascript/agenda-code.js
```diff
@@ -1346,52 +1346,55 @@ DynamicList.prototype.renderDatesHTML = function(rows, index) {
...
-  for (var i = 0; i < numberOfPlacholderDays; i++) { 
-    var newDate = {
-      week: moment(firstDate).utc().subtract(i + 1, 'days').format("ddd"),
-      day: moment(firstDate).utc().subtract(i + 1, 'days').format("DD"),
-      month: moment(firstDate).utc().subtract(i + 1, 'days').format("MMM"),
-      placeholder: true
+
+  if (clonedRecords.length) {
+    // set first date in agenda
+    firstDate = new Date(clonedRecords[0].data[dateField]).toUTCString();
```
angelnikolov: Is there a case where `clonedRecords[0]` will be `null?`

js/layout-javascript/agenda-code.js
```diff
@@ -1346,52 +1346,55 @@ DynamicList.prototype.renderDatesHTML = function(rows, index) {
...
+
+    // set last date in agenda
+    lastDate = new Date(clonedRecords[clonedRecords.length - 1].data[dateField]).toUTCString();
+
+    // Adds 5 days before the first date
+    // Save them in an array
+    for (var i = 0; i < numberOfPlacholderDays; i++) {
```
angelnikolov: Typo
```suggestion
    for (var i = 0; i < numberOfPlaceholderDays; i++) { 
```

js/layout-javascript/agenda-code.js
```diff
@@ -1653,6 +1656,10 @@ DynamicList.prototype.centerDate = function() {
   var activeWidth = _this.$container.find('.agenda-date-selector li.active').outerWidth();
   var halfWidth = activeWidth / 2;
 
+  if (!activePosition || isNaN(activePosition.left) || isNaN(halfWindowWidth) || isNaN(halfWidth)) {
```
angelnikolov: What else can `activePosition.left`, `halfWindowWidth` and `halfWidth` be instead of a number?

js/layout-javascript/agenda-code.js
hcarneiro: I don't think so. A data source API returns an array, even when empty.

js/layout-javascript/agenda-code.js
hcarneiro: If it's not a number we can't use it, hence why that if statement.

I just care that it is a number.

hcarneiro: Any concerns?
I know that if it is `null` it will give a false positive.

But from my tests it will never be `null` it can either be a `Number` or `undefined`

angelnikolov: Those were my concerns and that's why I asked but you summed it up. Thanks!

PR #111, opened 2019-02-08
Added a query parser to LFD

angelnikolov:
This PR adds a query parser to all LFD layouts.
I've followed a similar naming convention to the one for Storage.
These are some examples of queries which will work the same way as if they were applied through PV (i.e Storage):
`dynamicListOpenId=1128` - opens a specific entry

`dynamicListSearchColumn=Name&dynamicListSearchValue=Angel` - will search for `Angel` in the column `Name`

`dynamicListFilterValue=Angel,28` - will select the filter options `Angel` and `28`

`dynamicListFilterValue=Angel,28&dynamicListFilterHideControls=true` - will select the filter options `Angel` and `28` **and will also hide the controls**

`dynamicListPrefilterColumn=Name&dynamicListPrefilterValue=Angel&dynamicListPrefilterLogic=contains` - will pre-filter the list, by testing if any `Name` `contains` `Angel`

`dynamicListPrefilterColumn=Name,Age&dynamicListPrefilterValue=Younger,29&dynamicListPrefilterLogic=contains,%3C` - will pre-filter the list, by testing if any `Name` `contains` `Angel` AND `Age` `is lower` than `29`

**NOTE**
I intentionally left out two callbacks which could be passed to the LFD through storage.
Both `goBack.action` and  `previousScreen` are evaluated in the browser of the user and while we are certain that no malicious code could have been inserted in the Storage, because a user having access to the project must have added those callbacks there, we can't be sure that someone won't click on a shortened url, containing a function in the url which once executed - could harm the user in some way.

### DIFF
--- a/build.html
+++ b/build.html
@@ -37,25 +37,23 @@
       };
     </script>
     <div data-script-insert="{{ id }}" class="hidden"></div>
-    <script type="text/javascript" src="{{{ asset "js/query-parser.js" }}}"></script>
     <script type="text/javascript">
-      var script = document.createElement('script');
-      script.onload = function() {
-        if (typeof DynamicList !== 'function') {
-          console.warn('DynamicList() is not defined for widget for widget ID {{ id }}');
-          return;
-        }
-
-        var id = {{ id }};
-        DynamicListFuncs[id] = DynamicList;
-        DynamicList = null;
-        Fliplet().then(function () {
+      Fliplet().then(function(){
+        var script = document.createElement('script');
+        script.onload = function() {
+          if (typeof DynamicList !== 'function') {
+            console.warn('DynamicList() is not defined for widget for widget ID {{ id }}');
+            return;
+          }
+          var id = {{ id }};
+          DynamicListFuncs[id] = DynamicList;
+          DynamicList = null;
           var data = Fliplet.Widget.getData(id);
           dynamicLists[id] = new DynamicListFuncs[id](id, data);
-        })
-      };
-      script.src = assetUrl['{{ layout }}'];
-      document.querySelector('[data-script-insert="{{ id }}"]').appendChild(script);
+        };
+        script.src = assetUrl['{{ layout }}'];
+        document.querySelector('[data-script-insert="{{ id }}"]').appendChild(script);
+      });
     </script>
   {{/if}}
 {{else}}
--- a/build.html
+++ b/build.html
@@ -37,6 +37,7 @@
       };
     </script>
     <div data-script-insert="{{ id }}" class="hidden"></div>
+    <script type="text/javascript" src="{{{ asset "js/query-parser.js" }}}"></script>
     <script type="text/javascript">
       var script = document.createElement('script');
       script.onload = function() {
--- a/js/query-parser.js
+++ b/js/query-parser.js
@@ -5,7 +5,7 @@
  */
 Fliplet.Registry.set('dynamicListQueryParser', function() {
 	// we do not execute previousScreen like in the PV case so we don't open ourselves up to an xss attack
-	this.previousScreen =Fliplet.Navigate.query['dynamicListPreviousScreen'] === 'true';
+	this.previousScreen = Fliplet.Navigate.query['dynamicListPreviousScreen'] === 'true';
 
 	// action is intentionally ommited so we don't open ourselves up to an xss attack
 	this.pvGoBack = _.pickBy({
@@ -98,7 +98,7 @@ Fliplet.Registry.set('dynamicListQueryParser', function() {
 		this.pvFilterQuery.value = filterValues && filterValues.length ? filterValues : this.pvFilterQuery.column;
 
 		// cast to boolean
-    this.pvFilterQuery.hideControls = this.pvFilterQuery.hideControls === 'true';
+		this.pvFilterQuery.hideControls = this.pvFilterQuery.hideControls === 'true';
 		this.data.filtersEnabled = this.queryFilter;
 	} else {
 		this.queryFilter = null;
--- a/js/query-parser.js
+++ b/js/query-parser.js
@@ -4,111 +4,111 @@
  * for prepopulating, prefiltering and opening an entry
  */
 Fliplet.Registry.set('dynamicListQueryParser', function() {
-	// we do not execute previousScreen like in the PV case so we don't open ourselves up to an xss attack
-	this.previousScreen = Fliplet.Navigate.query['dynamicListPreviousScreen'] === 'true';
+  // we do not execute previousScreen like in the PV case so we don't open ourselves up to an xss attack
+  this.previousScreen = Fliplet.Navigate.query['dynamicListPreviousScreen'] === 'true';
 
-	// action is intentionally ommited so we don't open ourselves up to an xss attack
-	this.pvGoBack = _.pickBy({
-		enableButton: Fliplet.Navigate.query['dynamicListEnableButton'],
-		hijackBack: Fliplet.Navigate.query['dynamicListHijackBack']
-	});
-	this.queryGoBack = _(this.pvGoBack).size() > 0;
+  // action is intentionally ommited so we don't open ourselves up to an xss attack
+  this.pvGoBack = _.pickBy({
+    enableButton: Fliplet.Navigate.query['dynamicListEnableButton'],
+    hijackBack: Fliplet.Navigate.query['dynamicListHijackBack']
+  });
+  this.queryGoBack = _(this.pvGoBack).size() > 0;
 
-	// cast to booleans
-	this.pvGoBack.enableButton = this.pvGoBack.enableButton === 'true';
-	this.pvGoBack.hijackBack = this.pvGoBack.hijackBack === 'true';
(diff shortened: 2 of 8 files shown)


js/query-parser.js
```diff
@@ -0,0 +1,118 @@
+/**
```
squallstar: indentation is 2 spaces

build.html
```diff
@@ -37,6 +37,7 @@
       };
     </script>
     <div data-script-insert="{{ id }}" class="hidden"></div>
+    <script type="text/javascript" src="{{{ asset "js/query-parser.js" }}}"></script>
```
squallstar: Nope. This asset should be listed in the widget.json configuration!

tonytlwu:
Adding https://docs.google.com/spreadsheets/d/1XCH3K2lehm0IO9LpmBo5sU1b1Fhd2TkvxuN2L0qOTas/edit#gid=0 for reference.

1. Good PR description explainer.
1. Good structure for chaining conditions

For this filter option:

```
dynamicListFilterValue=Angel,28
```

Which column will the filter be run from? It doesn't seem to be clear whether it's `First Name === Angel` or `Location === Angel` or any other columns.

build.html
tonytlwu: Doesn't this mean the file will be loaded multiple times if there are multiple instances on the page?

squallstar: Yep :) I've commented too that the file should be listed in the widget.json

js/query-parser.js
```diff
@@ -0,0 +1,118 @@
...
+	} else {
+		this.queryFilter = null;
+	}
+	return (
+		this.previousScreen ||
+		this.queryGoBack ||
+		this.queryPreFilter ||
+		this.queryOpen ||
+		this.querySearch ||
+		this.queryFilter
```
tonytlwu: I don't know if there's a right/wrong or a preferred way for Fliplet to do this (cc @squallstar ) but I usually go for

```js
		this.previousScreen
  		|| this.queryGoBack
  		|| this.queryPreFilter
  		|| this.queryOpen
  		|| this.querySearch
  		|| this.queryFilter
```

...with:

1. The operator in the beginning of the line
1. Indenting the subsequent lines

js/query-parser.js
```diff
@@ -0,0 +1,118 @@
...
+		hideControls: Fliplet.Navigate.query['dynamicListFilterHideControls']
+	});
+	this.queryFilter = _(this.pvFilterQuery).size() > 0;
+	if (this.queryFilter) {
+		// check if a comma separated list of columns were passed as column
+		var filterValues = this.pvFilterQuery.value ? this.pvFilterQuery.value.split(',') : null;
+		this.pvFilterQuery.value = filterValues && filterValues.length ? filterValues : this.pvFilterQuery.column;
+
+		// cast to boolean
+		(this.pvFilterQuery.hideControls = this.pvFilterQuery.hideControls
```
tonytlwu: Is this not just...

```suggestion
		(this.pvFilterQuery.hideControls = this.pvFilterQuery.hideControls
```

build.html
angelnikolov: I tried, but for some reason the LFD got loaded before it.. Will try again.

angelnikolov:
> Adding https://docs.google.com/spreadsheets/d/1XCH3K2lehm0IO9LpmBo5sU1b1Fhd2TkvxuN2L0qOTas/edit#gid=0 for reference.
> 
> 1. Good PR description explainer.
> 2. Good structure for chaining conditions
> 
> For this filter option:
> 
> ```
> dynamicListFilterValue=Angel,28
> ```
> Which column will the filter be run from? It doesn't seem to be clear whether it's `First Name === Angel` or `Location === Angel` or any other columns.

It will be run off based on order. I am open for a better idea, but since we can't do nesting in the url we have just two options:
1. Custom naming convention - i.e `dynamicListFilterValue1=Angel&dynamicListFilterValue2=Tony`
2. Order based naming convention - what I've done.

build.html
squallstar: @angelnikolov did you try adding it as first asset in the list?

angelnikolov: Yes, as this is true for all of the LFD code anyway. This is just a way of abstracting the parser out and not having to repeat in every single layout code.

tonytlwu:
> > Adding https://docs.google.com/spreadsheets/d/1XCH3K2lehm0IO9LpmBo5sU1b1Fhd2TkvxuN2L0qOTas/edit#gid=0 for reference.
> > 
> > 1. Good PR description explainer.
> > 2. Good structure for chaining conditions
> > 
> > For this filter option:
> > ```
> > dynamicListFilterValue=Angel,28
> > ```
> > Which column will the filter be run from? It doesn't seem to be clear whether it's `First Name === Angel` or `Location === Angel` or any other columns.
> 
> It will be run off based on order. I am open for a better idea, but since we can't do nesting in the url we have just two options:
> 
> 1. Custom naming convention - i.e `dynamicListFilterValue1=Angel&dynamicListFilterValue2=Tony`
> 2. Order based naming convention - what I've done.

It's not the order I'm talking about. In a collection such as below:

```js
[
  { name: 'Tony', age: 28, yearsAtFliplet: 3 },
  { name: 'Angel', age: 53, yearsAtFliplet: 28 }
]
```

When I run the filter `dynamicListFilterValue=Angel,28`, is the filter based on the `age` or `yearsAtFliplet` property? It should be specific to a column. You had the for `dynamicListPrefilterColumn=Name&dynamicListPrefilterValue=Angel&dynamicListPrefilterLogic=contains`. I'm not sure why you didn't go for `dynamicListFilterColumn=Name&dynamicListFilterValue=Angel&dynamicListFilterLogic=contains`.

build.html
angelnikolov: Of course, the issue isn't with that. Adding it as the first element in the list will put in @ the top of the `rendering.js` file for the LFD. However, the way the lfd code is inserted on the page is hardcoded in `build.html` file here as
```
    <div data-script-insert="{{ id }}" class="hidden"></div>
```
which later transpiles to `https://d2fx5g1ro9ok5z.cloudfront.net/widgets/com.fliplet.dynamic-lists/1.4.0/js/layout-javascript/small-card-code.js?_=1549623239` from Handlebars.
By the time this loads, the `rendering.js` file has not even been requested by `Fliplet.require` since that call is lower in the dom..

angelnikolov:
`dynamicListFilterColumn=Name&dynamicListFilterValue=Angel&dynamicListFilterLogic=contains` this will work too, not sure what you meant by "didn't go for". It is working.
However, what if you want to filter by `Angel` contained `Name` and by lower than `28` for `Age`? This is currently supported by the storage way of filtering.
That's why I did it in an order-based fashion, like:
`dynamicListFilterColumn=Name,Age&dynamicListFilterValue=Angel,28&dynamicListFilterLogic=contains, <`
So this will create one filter for the first set of values:
```
{
    column: 'Name',
    value: 'Angel',
    logic: 'contains
}
```
and another one for the second set:

```
{
    column: 'Age',
    value: 28,
    logic: '<'
}
```

tonytlwu:
@angelnikolov I didn't know that you implemented it because it wasn't in the PR description. Did I misread it?

js/query-parser.js
angelnikolov: Absolutely! Thanks

angelnikolov:
Actually, this is just for prefiltering I misread prefilter as filter, sorry.
Back to the original question - filtering only takes values, not columns as per the documentation here:
https://developers.fliplet.com/API/components/list-from-data-source.html
If you want it to also target specific columns, as the prefiltering example from my previous comments we will need changes to the way the passed filters are applied from both Storage and Query.

build.html
squallstar: I see. A few questions for you and @hcarneiro:
1) why is the script URL being printed out in the DOM? Is it because you only include a JS file to the selected templated without having to include all of them? If so, I assume you parse this somewhere in your code and inject the script as a script tag, correct? @angelnikolov given that, you just need to load the query-parser before/after such file (depending on what you need)

angelnikolov: That is correct, and that's exactly what the code is doing, by adding the query-parser before the file. Adding it to the assets in widget.json wouldn't work in this case, because it will always be after.
However I just thought of another way to load the specific layout's code and that's to just wrap the whole script loading in `Fliplet.then()`, please review it again.

js/query-parser.js
angelnikolov: For some reason all new files that I create are formatted by tabs, even though the IDE shows that it's spaces.. I need to change it to tabs, and then change it back to spaces to work.. Weird

tonytlwu:
> Actually, this is just for prefiltering I misread prefilter as filter, sorry.
> Back to the original question - filtering only takes values, not columns as per the documentation here:
> https://developers.fliplet.com/API/components/list-from-data-source.html
> If you want it to also target specific columns, as the prefiltering example from my previous comments we will need changes to the way the passed filters are applied from both Storage and Query.

@angelnikolov I see. I think it's because of [this filter issue](https://github.com/Fliplet/fliplet-studio/issues/3761). We will need to fix it before we can have column-specific filters. It's fine for now then.

PR #119, opened 2019-03-06
Fixes bookmark button position when card has no image

hcarneiro:
Issue ref: https://github.com/Fliplet/fliplet-studio/issues/3995

screenshot 2019-03-06 at 10 21 58
screenshot 2019-03-06 at 10 22 06


squallstar:
@hcarneiro in your avatar in the first screenshot you look different... you changed a bit!

PR #132, opened 2019-03-20
Fix for phones with notch

hcarneiro:
We use the value of the top margin to figure out the top offset for the expand animation, and because the HTML tag has a `margin-top` with `env(safe-area-inset-top)` we can also use this value for phones with notches.

### DIFF
--- a/js/layout-javascript/small-card-code.js
+++ b/js/layout-javascript/small-card-code.js
@@ -2258,7 +2258,7 @@ DynamicList.prototype.expandElement = function(elementToExpand, id) {
 
     // Get the size of the html offset
     // This get into account phones with notch
-    var computedStyles = window.getComputedStyle(document.getElementsByTagName("html")[0]);
+    var computedStyles = window.getComputedStyle(document.getElementsByTagName('html')[0]);
     var htmlMarginTop = computedStyles.getPropertyValue('margin-top');
     var toTop = parseInt(htmlMarginTop.replace('px',''), 10);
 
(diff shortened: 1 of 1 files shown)


js/layout-javascript/small-card-code.js
```diff
@@ -2256,6 +2256,12 @@ DynamicList.prototype.expandElement = function(elementToExpand, id) {
     var directoryDetailImageWrapper = elementToExpand.find('.small-card-list-detail-image-wrapper');
     var directoryDetailImage = elementToExpand.find('.small-card-list-detail-image');
 
+    // Get the size of the html offset
+    // This get into account phones with notch
+    var computedStyles = window.getComputedStyle(document.getElementsByTagName("html")[0]);
```
tonytlwu:
```suggestion
    var computedStyles = window.getComputedStyle(document.getElementsByTagName('html')[0]);
```

js/layout-javascript/small-card-code.js
```diff
@@ -2256,6 +2256,12 @@ DynamicList.prototype.expandElement = function(elementToExpand, id) {
...
+    var htmlMarginTop = computedStyles.getPropertyValue('margin-top');
+    var toTop = parseInt(htmlMarginTop.replace('px',''), 10);
```
tonytlwu: The `.replace()` shouldn't be necessary.

```suggestion
    var toTop = parseInt(htmlMarginTop, 10);
```

image

PR #133, opened 2019-03-24
Changes for LFD to support building podcast app

tonytlwu:
- Adds new `flListDataAfterGetData` hook where `data.records` contains records that can be manipulated.
- Tweaks news feed card styles to ensure rounded corners are clipped in Safari and iOS apps.
- Tweaks filter parsing to allow `null` attributes.

### DIFF
--- a/css/layout-css/news-feed-style.upload.css
+++ b/css/layout-css/news-feed-style.upload.css
@@ -558,6 +558,8 @@
   opacity: 1;
   margin-bottom: 40px;
   transition: box-shadow 0.15s ease-out;
+  -webkit-transform: translate3d(0, 0, 0);
+  transform: translate3d(0, 0, 0);
 }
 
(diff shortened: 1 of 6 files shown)


css/layout-css/news-feed-style.upload.css
```diff
@@ -558,6 +558,8 @@
   opacity: 1;
   margin-bottom: 40px;
   transition: box-shadow 0.15s ease-out;
+  -webkit-transform: translate3d(0, 0, 0);
```
hcarneiro: @tonytlwu Did you test this change on:
1- One or two layouts from production? (copying the HTML from a layout on production)
2- Inside an accordion?
3- Expanding and collapsing the card details?

tonytlwu: I haven’t tested the first one. Let me give that a go 👍🏼

PR #137, opened 2019-03-29
Move common data processing functionality to a utility library

tonytlwu:
Refactors the following:

- `DynamicList.Utils.Record.contains(record, value)` used in searches to check if a record contains a search term
- `DynamicList.Utils.Records.runFilters(records, filters)` used to filter records based on a set of filters configured in Studio
- `DynamicList.Utils.Date.moment(date)` used to create a date based on a timestamp, taking timezones into consideration
- `DynamicList.Utils.Records.getFields(records, key)` used to get the common list of properties across the entire collection. Optionally provide a `key` for caching.
- Centralizes Handlebars helpers to use `DynamicList.Utils.registerHandlebarsHelpers()`


squallstar:
@tonytlwu you mentioned `Fliplet.Utils`, I assume that's a mistake?

tonytlwu:
@squallstar Yup, a mistake

js/utils.js
```diff
@@ -0,0 +1,23 @@
+Fliplet.Registry.set('dynamicListUtils', function() {
+  function dataContains(data, value) {
+    if (!data) {
+      return false;
+    }
+
+    if (_.isArray(data)) {
+      return _.some(data, function (el) {
+        return el.toLowerCase().indexOf(value) > -1;
```
squallstar: what if `el` is not a string?

squallstar: here you should have called the same function in a recursive way instead of redefining the logic

tonytlwu: https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/137/commits/45ea25a11add4479bb11d17c48901f26c8650e32 and now we don't have to apply the fix 6 times ☺️

js/utils.js
```diff
@@ -0,0 +1,23 @@
+Fliplet.Registry.set('dynamicListUtils', function() {
```
squallstar: `dynamicListSharedLibraries`?
`dynamicListHelpers`?

js/utils.js
squallstar: @tonytlwu but you're defining the same logic twice in the same function. Definitely wouldn't pass code quality standards :)

squallstar: As I suggested, recursive is the preferred way to handle this situation so you avoid duplicate logic and also deal with nested arrays.

tonytlwu: https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/137/commits/97dd0a9c0077568f3e5fc7cdd0e992bbc863de3f

PR #152, opened 2019-04-30
WIP - Include fliplet-icons

hcarneiro:
This will fix an issue where the fliplet icons files were missing in the component.
We want the fliplet icons library included in the component mainly for us to be able to use them in custom builds for clients.


hcarneiro:
This is unnecessary

PR #156, opened 2019-05-03
Make sure filter logic is rendered consistently when loading/adding filters

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/4263

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -183,11 +183,16 @@ Fliplet.Registry.set('dynamicListUtils', function() {
 
     return _.filter(records, function(record) {
       return _.every(filters, function(filter) {
+        if (!filter.condition || !filter.value || filter.column === 'none') {
+          // Filter isn't configured correctly
+          return true;
+        }
+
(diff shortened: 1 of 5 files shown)


js/utils.js
```diff
@@ -183,11 +183,16 @@ Fliplet.Registry.set('dynamicListUtils', function() {
 
     return _.filter(records, function(record) {
       return _.every(filters, function(filter) {
+        if (!filter.condition || !filter.value || filter.column === 'none') {
```
squallstar: what if one of these is `undefined`, can it happen? e.g. `filters [ undefined, null ]`

tonytlwu: It can't. The default is created as `"none"` (just pushed a commit to match this), and the `<select>` has a default value of `"none'` as well. I don't like it and would've preferred `null` but I'm not risking changing this.

tonytlwu:
@squallstar All tested (now that I'm able to `fliplet publish`). Ready for review.

PR #166, opened 2019-05-14
Fixes issue where multiple filter fields cannot share the same value

tonytlwu:
Merging into `feat/filter-rebuild` for https://github.com/Fliplet/fliplet-studio/issues/4343

Ref. https://github.com/Fliplet/fliplet-studio/issues/3761


js/layout-javascript/small-card-code.js
```diff
@@ -1433,92 +1431,81 @@ DynamicList.prototype.getPermissions = function(entries) {
...
+    // Legacy class-based filters
+    $activeFilterControls.each(function(index, element) {
+      _this.filterClasses.push($(element).data('toggle'));
     });
 
-    // If "_.includes" returns TRUE
-    // we actually want to return FALSE to _.filter
-    return !_.includes(matched, false);
-  });
+    filteredData = _.filter(listData, function(row) {
```
squallstar: is there any chance these functions can be put into utilities so they're not repeated for each layout?

tonytlwu: I think so, yes. Let me do that.

tonytlwu: @squallstar Refactored

PR #171, opened 2019-05-23
Ensure users can add/edit/delete own entries as configured

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/4402

Also implements the following from https://github.com/Fliplet/fliplet-studio/issues/4371

> LFD is updated to verify admins by checking for arrays when verifying admin status


hcarneiro:
❤️ Amazing refactor here!

PR #176, opened 2019-05-30
Allow URLs as images without needing any files in the selected folder

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/4449

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -297,7 +297,7 @@ Fliplet.Registry.set('dynamicListUtils', function() {
     return Promise.resolve(fields);
   }
 
-  function connectToGetFiles(data) {
+  function getFiles(data) {
     var cacheKey = JSON.stringify(data.query);
 
     if (!cachedFiles[cacheKey]) {
@@ -326,7 +326,7 @@ Fliplet.Registry.set('dynamicListUtils', function() {
         var column = data.field.column;
 
         if (isValidImageUrl(entryData[column])) {
-          // Entry data doesn't need updating
+          // Record data doesn't need updating
           return data.record;
         }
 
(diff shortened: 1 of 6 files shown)


js/utils.js
```diff
@@ -285,6 +297,180 @@ Fliplet.Registry.set('dynamicListUtils', function() {
...
+
+      if (!forComments) {
+        _.forEach([config['summary-fields'], config.detailViewOptions], function (fields) {
+          _.forEach(fields, function(field) {
+            if (field.type !== 'image') {
+              return;
+            }
+
+            switch (field.imageField) {
+              case 'app':
```
hcarneiro: this `case 'app'` and the `case 'organization'` can probably be removed since you removed the options in the dropdown.

js/utils.js
```diff
@@ -285,6 +297,180 @@ Fliplet.Registry.set('dynamicListUtils', function() {
...
+                }
+                break;
+              default:
+                break;
+            }
+          });
+        });
+      } else {
+        switch (config.userFolderOption) {
+          case 'app':
```
hcarneiro: Same here!

js/utils.js
tonytlwu: The app setting currently exists in widget instances, so I think the build output needs to support the behavior.

js/utils.js
tonytlwu: As above

js/utils.js
```diff
@@ -339,7 +525,8 @@ Fliplet.Registry.set('dynamicListUtils', function() {
     },
     Records: {
       runFilters: runRecordFilters,
-      getFields: getRecordFields
+      getFields: getRecordFields,
```
hcarneiro: @tonytlwu `getRecordFields` does not exist anywhere in these changes.

tonytlwu: @hcarneiro It's [here](https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/f34ea68aa7d09fe9a98496e4378667e4fb2cd439/js/utils.js#L279).

It's not in the changes because no changes were made to the function. I only added a `,` at the end.

PR #191, opened 2019-06-20
Use "dynamicListOpenSingleEntry" to open an entry if search results only return one result

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/4563

### DIFF
--- a/css/layout-css/news-feed-style.upload.css
+++ b/css/layout-css/news-feed-style.upload.css
@@ -412,6 +412,14 @@
   line-height: 1;
 }
 
+.new-news-feed-list-container .current-query {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: calc(100% - 185px);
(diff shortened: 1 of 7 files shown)


css/layout-css/news-feed-style.upload.css
```diff
@@ -412,6 +412,14 @@
   line-height: 1;
 }
 
+.new-news-feed-list-container .current-query {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: calc(100% - 185px);
```
squallstar: what's 185px?

tonytlwu: It's the space for the "Searching for:" label on the left, and a bit more (for buffer).

image

Ideally, I'd change the HTML structure a bit to avoid needing to code in a number like this, but I'd like to get this fix QA tested for a deployment on Monday.

PR #200, opened 2019-07-24
The "Replace folder" button is not aligned when the folder is chosen

MaksymShokin:
@squallstar @tonytlwu 

**Fixed behavior:**
Fixed issue that the "Replace folder" button was not aligned when the folder is chosen.
Added max-height for image-selector.

ref https://github.com/Fliplet/fliplet-studio/issues/4215

replace-button-fix-demo (online-video-cutter com)


tonytlwu:
Now that the changes have been made, it look like the UI no longer tells the user what the selected folder is, like it used to:

image

A better fix would be to resolve it such that it looks like this:

image

image

MaksymShokin:
It does show the selected folder. 
replace button img

tonytlwu:
Excellent 👍 I didn't see it in the original screencast posted.

MaksymShokin:
This is how it should be?

Replace button q1

tonytlwu:
@MaksymShokin No, I don't think it should be truncated. It should behave like this.

image

And basically your solution of using max-height can't work. The container needs to expand and push other content down as normal UI flow does.

MaksymShokin:
Now works as expected.

Replace button fix demo 3

tonytlwu:
@MaksymShokin I see that the labels **Primary text**, **Secondary text** etc. have all been updated to be right aligned. This should not happen. As you can see in the image below, it's only the **Data type**, **[Custom field](https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/dc8012faac8d5d500d7d3650dee0dad89452185f/templates/interface/detail-view-panels.interface.hbs#L46-L48)** and **[Custom label](https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/dc8012faac8d5d500d7d3650dee0dad89452185f/templates/interface/detail-view-panels.interface.hbs#L74-L76)** labels that should be right aligned.

replace button img

MaksymShokin:
@tonytlwu Fixed right-alignment problem

tonytlwu:
@MaksymShokin Please resolve conflict and re-test.

MaksymShokin:
@tonytlwu Conflicts resolved

PR #207, opened 2019-07-29
Prevent configuration loss if the interface is saved before everything has loaded

MaksymShokin:
@squallstar @tonytlwu 

**Fixed behavior**:
When clicking on the Save & Close button before all data is loaded, the component is saving empty data. 

Added variable _this.isLoaded which is false by default, and its value changes to true when data is loaded. Added check to Fliplet.Widget.onSaveRequest which is allowing to save component only when data is loaded.

ref https://github.com/Fliplet/fliplet-studio/issues/4370

save-before-load

### DIFF
--- a/js/interface-lists.js
+++ b/js/interface-lists.js
@@ -512,6 +512,7 @@ function attahObservers() {
   Fliplet.Widget.onSaveRequest(function () {
     if (!dynamicLists.isLoaded) {
       Fliplet.Widget.complete();
+      return;
     }
     $('form').submit();
   });
--- a/js/interface-lists.js
+++ b/js/interface-lists.js
@@ -282,7 +282,7 @@ function attahObservers() {
             value: widgetData.userAdminColumn,
             field: '#select_user_admin'
           });
-          
+
           values.forEach(function(field) {
             if (!validate(field.value)) {
               errors.push(field.field);
@@ -330,7 +330,7 @@ function attahObservers() {
           if (!widgetData.userNameFields && !widgetData.userNameFields.length) {
             errors.push('#user-name-column-fields-tokenfield');
           }
-          
+
           values.forEach(function(field) {
             if (!validate(field.value)) {
               errors.push(field.field);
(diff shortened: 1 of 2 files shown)


js/interface-lists.js
```diff
@@ -510,7 +510,9 @@ function attahObservers() {
   });
 
   Fliplet.Widget.onSaveRequest(function () {
-    $('form').submit();
+    if (dynamicLists.isLoaded) {
+      $('form').submit();
+    }
```
tonytlwu: @MaksymShokin Looks like you've managed to find the problem and the right point to make changes. However, the method you've used results a **Save & Close** button that doesn't do anything when users click on it. Users would feel that something isn't working.

Instead, I'd recommend running `Fliplet.Widget.complete();` if `!dynamicLists.isLoaded`, which will close the widget interface without saving anything.

MaksymShokin: @tonytlwu Fixed. If everything is fine could you please merge this PR because it holds the PR from this task https://github.com/Fliplet/fliplet-studio/issues/4660.

js/interface-lists.js
```diff
@@ -510,9 +510,10 @@ function attahObservers() {
   });
 
   Fliplet.Widget.onSaveRequest(function () {
-    if (dynamicLists.isLoaded) {
-      $('form').submit();
+    if (!dynamicLists.isLoaded) {
+      Fliplet.Widget.complete();
```
tonytlwu: @maksim-an-upplabs Can you explain why you haven't stopped the code execution with `return` after this? Otherwise, it looks like if the configuration isn't loaded, it will still run `$('form').submit()`, which would replicate the reported bug.

tonytlwu: @MaksymShokin Can you explain why you haven't stopped the code execution with `return` after this? Otherwise, it looks like if the configuration isn't loaded, it will still run `$('form').submit()`, which would replicate the reported bug.

MaksymShokin: @tonytlwu When I was testing this code on staging it worked fine the way it is. If you insist I can add an extra return to be safe.

tonytlwu: @MaksymShokin I imagine it's probably because `Fliplet.Widget.complete()` is synchronous and closes the pane immediately. Let's add a `return` to be clearer for future developers that the execution should be stopped there.

MaksymShokin: @tonytlwu Fixed

PR #210, opened 2019-07-31
Updated entry limit message to reflect the enabled search and filter features

YaroslavOvdii:
@squallstar @tonytlwu 

Fixed behavior:
When we limit the number of entries and not using search or filter the message displayed was incorrect.

Remove text creation from the template and made a function to create a message.

ref https://github.com/Fliplet/fliplet-studio/issues/4422

message-on-limited-entries


js/interface.js
```diff
@@ -2371,6 +2371,7 @@ var DynamicLists = (function() {
       if (limit && limit.length && /^\d+$/.test(limit)) {
         data.enabledLimitEntries = true;
         data.limitEntries = parseInt(limit, 10);
+        data.limitEntriesText = constructEntriesText(data);
```
tonytlwu: I don't like that the copy has gone from a template file, which is updateable through a widget update deployment, to being saved in the widget configuration, which is only updateable if users open and save the widget configuration in Studio.

YaroslavOvdii: Then it would be better to return this solution to the template file and remove from widget configuration, right?

tonytlwu: Correct

YaroslavOvdii: @tonytlwu  Return solution to a template.

templates/build/news-feed-base.build.hbs
```diff
@@ -103,7 +103,7 @@
       \{{/ifCond}}
     \{{/ifCond}}
     \{{#ifCond limitEntries '>' 0}}
-    <p>Showing the first \{{limitEntries}} list items. \{{#if searchEnabled}}Search\{{/if}}\{{#ifCond searchEnabled '&&' filtersEnabled}} and/or \{{/ifCond}}\{{#if filtersEnabled}}Filter\{{/if}} to see more.</p>
+    <p>Showing the first \{{limitEntries}} \{{#ifCond searchEnabled '||' filtersEnabled}} list items. \{{#if searchEnabled}}Search\{{/if}}\{{#ifCond searchEnabled '&&' filtersEnabled}} and/or \{{/ifCond}}\{{#if filtersEnabled}}Filter\{{/if}} to see more. \{{else}} lists.\{{/ifCond}}</p>
```
tonytlwu: 1. It looks like if neither search or filter is enabled, the message would say "Showing the first {{limitEntries}} lists.". Am I right? If so, it should be "Showing the first {{limitEntries}} list items."

YaroslavOvdii: @tonytlwu Fix this.

PR #211, opened 2019-07-31
Horizontal cards are not center-aligned on desktop and tablet screens

YaroslavOvdii:
@tonytlwu @squallstar 

Fixed behavior:
Horizontal cards weren't center-aligned on desktop and tablet screens.

Added justify-content center for the component main block.

ref https://github.com/Fliplet/fliplet-studio/issues/4213

Horizontal-cards-wasn_t-center-aligned


tonytlwu:
No longer necessary as commented in https://github.com/Fliplet/fliplet-studio/issues/4213#issuecomment-516928064

PR #213, opened 2019-07-31
allow link action to read from array; refactor to utils

squallstar:
ref https://github.com/Fliplet/fliplet-studio/issues/4769

---

Tested.

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -1140,6 +1140,28 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     });
   }
 
+  function openLinkAction(options) {
+    var entry = _.find(options.records, function(entry) {
+      return entry.id === options.recordId;
+    });
+
+    if (!entry) {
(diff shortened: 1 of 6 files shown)


js/utils.js
```diff
@@ -1140,6 +1140,28 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+  function openLinkAction(options) {
+    var entry = _.find(options.records, function(entry) {
+      return entry.id === options.recordId;
+    });
+
+    if (!entry) {
+      return;
+    }
+
+    var value = entry.data[options.summaryLinkAction.column];
```
tonytlwu: I'd also add tests for `options.summaryLinkAction.column` and `value` so that if `options.summaryLinkAction` isn't valid then `return` and do nothing.

squallstar: @tonytlwu done, please review

PR #232, opened 2019-08-28
Highlighting panel and 'Data view settings' button to show where error occurred

YaroslavOvdii:
@tonytlwu @squallstar 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/4877

## Description
Added error highlighting for panel and 'Data view settings' button

## Screenshots/screencasts
issue-4877

## Backward compatibility

### DIFF
--- a/js/interface-lists.js
+++ b/js/interface-lists.js
@@ -208,6 +208,19 @@ function validate(value) {
   return false;
 }
 
+function showOrHideError (showError, element) {
+  if (showError) {
+    var $element = $(element);
+    $element.addClass('has-error');
+    $element.parents('.form-group').addClass('has-error');
+    $element.parents('.panel').addClass('panel-danger').removeClass('panel-default');
+    return;
+  }
+    $('.has-error').removeClass('has-error');
+    $('.component-error').addClass('hidden');
+    $('.panel-danger').removeClass('panel-danger').addClass('panel-default');
+}
+
 function attahObservers() {
   $(document)
     .on('click', '[data-file-picker-user]', function() {
@@ -421,6 +434,23 @@ function attahObservers() {
             $('.component-error').addClass('hidden');
           }
         }
+        var imageFolderSelected = validateImageFoldersSelection();
+        if (imageFolderSelected) {
+          highlightError(selectedFieldId, false);
+
+          $('[data-relations-fields]').addClass('btn-default').removeClass('relations-error');
+        } else {
+          highlightError(selectedFieldId, true);
+
+          $('[data-relations-fields]').removeClass('btn-default').addClass('relations-error');
+
+          Fliplet.Modal.alert({
+            title: 'Invalid configuration',
+            message: 'Please review settings in <strong>Data view settings</strong> to continue.'
+          });
+
+          return;
+        }
 
         if (widgetData.pollEnabled && widgetData.pollColumn) {
           var errors = [];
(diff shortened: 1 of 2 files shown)


js/interface-lists.js
```diff
@@ -208,6 +208,19 @@ function validate(value) {
...
 
+function showOrHideError (showError, element) {
+  if (showError) {
+    var $element = $(element);
+    $element.addClass('has-error');
+    $element.parents('.form-group').addClass('has-error');
+    $element.parents('.panel').addClass('panel-danger').removeClass('panel-default');
+    return;
+  }
+    $('.has-error').removeClass('has-error');
```
tonytlwu: Unindent this section, and add a linebreak after the `if` clause

js/interface-lists.js
```diff
@@ -208,6 +208,19 @@ function validate(value) {
   return false;
 }
 
+function showOrHideError (showError, element) {
```
tonytlwu:
```suggestion
function toggleError (showError, element) {
```

js/interface-lists.js
```diff
@@ -452,8 +481,7 @@ function attahObservers() {
             }, 4000);
             return;
           } else {
-            $('.has-error').removeClass('has-error');
-            $('.component-error').addClass('hidden');
+            showOrHideError(false);
```
tonytlwu: I notice a `field` isn't provided here, but there's nothing in the function that deals with what happens if the second parameter isn't provided.

js/interface-lists.js
```diff
@@ -462,7 +490,7 @@ function attahObservers() {
           var values = [];
 
           values.push({
-            value: widgetData.pollColumn,
+            value: widgetData.surveyColumn,
```
tonytlwu: Why is this changed?

YaroslavOvdii: This is needed to avoid an error like this.
can-save-without-a-field

As you may see we able to save surveys without selecting any survey field. Because we were checking only the poll field.

js/interface-lists.js
YaroslavOvdii: We deal with this by if clause which fires when we provide a `true` as the first parameter and exit from the function by return in if clause. In another situation, we remove error classes by referring to them by classes.

js/interface-lists.js
YaroslavOvdii: @tonytlwu 
Renamed function

js/interface-lists.js
YaroslavOvdii: @tonytlwu 
Added a linebreak after `if` clause

tonytlwu: Also unindent this section, like below

```suggestion
  $('.has-error').removeClass('has-error');
```

js/interface-lists.js
```diff
@@ -421,6 +435,23 @@ function attahObservers() {
             $('.component-error').addClass('hidden');
           }
         }
+        var imageFolderSelected = validateImageFoldersSelection();
+        if (imageFolderSelected) {
```
tonytlwu: Line break before `if` clause

YaroslavOvdii: @tonytlwu 
Fixed

js/interface-lists.js
YaroslavOvdii: @tonytlwu 
Fixed

PR #234, opened 2019-08-29
Min height for lfd components 4881

maksim-an-upplabs:
@squallstar @tonytlwu 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/4881
## Description
Added css rules min-height for components in lfd widget to can see and edit it.
## Screenshots/screencasts
min-height-32px

## Backward compatibility


css/build.css
```diff
@@ -144,4 +144,8 @@
 /* For native devices */
 .native .fa-times-thin {
   padding-top: 4px;
-} 
\ No newline at end of file
+}
+
+.mode-interact .fl-widget-instance {
```
tonytlwu: I believe this would be applying a minimum height to all instances of all widgets. That's way too broad as a selector.

maksim-an-upplabs: @tonytlwu I've added new css class and rule for min-height.

css/build.css
```diff
@@ -144,4 +144,8 @@
 /* For native devices */
 .native .fa-times-thin {
   padding-top: 4px;
-} 
\ No newline at end of file
+}
+
+.dynamic-list-layout {
```
tonytlwu: You can just use

```suggestion
[data-dynamic-lists-id] {
```

I don't think a new class is necessary.

build.html
```diff
@@ -8,7 +8,7 @@
   {{{advancedSettings.cssCode}}}
   </style>
   {{/if}}
-  <div data-dynamic-lists-id="{{ id }}" data-dynamic-lists-layout="{{ layout }}" data-dynamic-lists-uuid="{{uuid}}"></div>
+  <div data-dynamic-lists-id="{{ id }}" data-dynamic-lists-layout="{{ layout }}" data-dynamic-lists-uuid="{{uuid}}" class="dynamic-list-layout"></div>
```
tonytlwu: What do we still need the `.dynamic-list-layout` class for?

```suggestion
  <div data-dynamic-lists-id="{{ id }}" data-dynamic-lists-layout="{{ layout }}" data-dynamic-lists-uuid="{{uuid}}"></div>
```

maksim-an-upplabs: @tonytlwu I'm sorry, I forgot to remove this class from html.

css/build.css
```diff
@@ -144,4 +144,8 @@
 /* For native devices */
 .native .fa-times-thin {
   padding-top: 4px;
-} 
\ No newline at end of file
+}
+
+[data-dynamic-lists-id] {
```
tonytlwu:
```suggestion
.mode-interact [data-dynamic-lists-id] {
```

The `.mode-interact` selector should be added so that the minimum height is only applied to edit mode. This is because we're only making this change to make sure the widget is editable. In preview, if the widget instance is empty, then it shouldn't take up unnecessary space with a minimum height.

tonytlwu:
@maksim-an-upplabs Looks like there are conflicts in `build.css`

maksim-an-upplabs:
> @maksim-an-upplabs Looks like there are conflicts in `build.css`
@tonytlwu Please check it.

css/build.css
```diff
@@ -148,4 +148,8 @@
 
 .native.ios .fa-times-thin {
   padding-top: 3px;
-} 
\ No newline at end of file
+}
+
+.mode-interact [data-dynamic-lists-id] {
+    min-height: 32px;
```
squallstar: two spaces, not four.

maksim-an-upplabs: fixed

PR #236, opened 2019-09-02
If bookmark is enabled added padding-right so text does not cover bookmark icon

MaksymShokin:
@tonytlwu @squallstar 

## Issue
Fliplet/fliplet-studio#4913

## Description
If bookmark if enabled class bookmark-padding will be added to div with all text.

## Screenshots/screencasts
Bookmark demo

## Backward compatibility


templates/build/small-card-loop.build.hbs
```diff
@@ -10,7 +10,7 @@
         \{{/if}}
       </div>
     </div>
-    <div class="small-card-list-text-wrapper">
+    <div class="small-card-list-text-wrapper \{{#if bookmarksEnabled}} bookmark-padding \{{/if}}">
```
squallstar: I would recommend calling this `has-bookmarks`, @tonytlwu do you agree?

squallstar: @tonytlwu ^

tonytlwu: I agree

MaksymShokin: @tonytlwu @squallstar Just to confirm in this case I will have to change the name from bookmarksEnabled to has-bookmarks though all the code. 
bookmarks demo

squallstar: @maksim-an-upplabs nope, what we meant is to change your `bookmark-padding` class to be named `has-bookmarks`.

MaksymShokin: @squallstar Fixed

PR #237, opened 2019-09-03
Preset configuration for agenda component 3188

maksim-an-upplabs:
@squallstar @tonytlwu 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/3188
## Description
Added functionality to set default value to HTML for content in agenda component.
## Screenshots/screencasts
https://cl.ly/d55bec92c24e
## Backward compatibility


js/interface.js
```diff
@@ -899,10 +899,10 @@ var DynamicLists = (function() {
                   id: index + 1,
                   columns: dataSourceColumns,
                   column: column,
-                  type: 'text',
+                  type: listLayout === 'agenda' ? 'html' : 'text',
```
tonytlwu: We would only want the content to be using the `html` field type. This change seems to risk all fields in the agenda layout to be added as a `html` type by default. That's too broad.

maksim-an-upplabs:
https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/237/commits/b83b68800107de2da3dd8064f7c8116ed13720cc - that changes should be more specific for content type 'html'.

tonytlwu:
@maksim-an-upplabs I don't think this will work. If I'm configuring an agenda with a data source where the agenda item details is found in a column called **Agenda content** instead of **Content**, then your code wouldn't work. You've specified it to only apply (and always apply) if the column is called **Content**, regardless of what the _intention_ is.

The solution should be to add a new detail view field when the default data is being loaded. They are defined in this file https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/48a985bcc82e62a9a73250cb97674f4d0641f2dc/js/default-configs/layouts-config.js#L202

I suggest you familiarise yourself with how this works.

maksim-an-upplabs:
> @maksim-an-upplabs I don't think this will work. If I'm configuring an agenda with a data source where the agenda item details is found in a column called **Agenda content** instead of **Content**, then your code wouldn't work. You've specified it to only apply (and always apply) if the column is called **Content**, regardless of what the _intention_ is.
> 
> The solution should be to add a new detail view field when the default data is being loaded. They are defined in this file
> 
> https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/48a985bcc82e62a9a73250cb97674f4d0641f2dc/js/default-configs/layouts-config.js#L202
> 
> I suggest you familiarise yourself with how this works.

@tonytlwu Can you tell me how I need to generate id fo default setting of content?

tonytlwu:
@maksim-an-upplabs The `id` can be randomly generated, as long as it's unique within the `detail-fields` array. Just make sure the `id` is the same number of characters as the other ones for consistency purposes.

maksim-an-upplabs:
@tonytlwu let me know if this changes is ok?)

PR #246, opened 2019-09-16
Validate and process date fields correctly in case it's not configured as the "Full Date" field

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/4989

### DIFF
--- a/js/layout-javascript/agenda-code.js
+++ b/js/layout-javascript/agenda-code.js
@@ -38,6 +38,7 @@ function DynamicList(id, data, container) {
   this.agendasByDay;
   this.searchedListItems;
   this.dataSourceColumns;
+  this.dateFieldLocation = 'Full Date';
 
   this.queryOpen = false;
   this.queryPreFilter = false;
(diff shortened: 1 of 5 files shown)


js/layout-javascript/agenda-code.js
```diff
@@ -38,6 +38,7 @@ function DynamicList(id, data, container) {
   this.agendasByDay;
   this.searchedListItems;
   this.dataSourceColumns;
+  this.dateFieldLocation = 'Full Date';
```
squallstar: dateFieldColumnName? `Location` is ambiguous to me

tonytlwu: @squallstar I imagine it'd be ambiguous (or just downright meaningless) to you because you don't know that there's a concept called `location` in the LFD. It's a predefined variable for **where** a field would be displayed. Search https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/master/js/default-configs/layouts-config.js for the word `location` and see.

PR #248, opened 2019-09-18
Added different padding's to Android and iOS native devices

YaroslavOvdii:
@tonytlwu @squallstar 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/4773

## Description
Added different padding's to Android and iOS devices

## Screenshots/screencasts
iOS 12
iOS closeIcon

Android 9
Screenshot_20190918-155336

## Backward compatibility

### DIFF
--- a/css/build.css
+++ b/css/build.css
@@ -146,6 +146,6 @@
   padding-top: 5px;
 }
 
-.native.iphonex .fa-times-thin {
+.native.ios .fa-times-thin {
   padding-top: 3px;
 } 
--- a/css/build.css
+++ b/css/build.css
@@ -142,6 +142,10 @@
 }
 
 /* For native devices */
-.native .fa-times-thin {
-  padding-top: 4px;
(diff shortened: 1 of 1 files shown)


css/build.css
```diff
@@ -142,6 +142,10 @@
 }
 
 /* For native devices */
-.native .fa-times-thin {
-  padding-top: 4px;
+.native.android .fa-times-thin {
+  padding-top: 5px;
+}
+
+.native.iphonex .fa-times-thin {
```
tonytlwu: Can you confirm that you're only adding this override to iPhone X (or iPhone X-like devices), and not to all iOS devices? I believe there's a `.native.ios` selector that you can use if you want to target all iOS devices.

YaroslavOvdii: @tonytlwu No, I was targeting all iOS devices. Sorry I didn't notice this class. I will make new PR to change it right away.

PR #254, opened 2019-10-07
Image data field now has field type set as image

MaksymShokin:
@tonytlwu @squallstar 

## Issue
Fliplet/fliplet-studio#4969

## Description
Added image data-field in image layouts.

## Screenshots/screencasts
Dynamic demo

## Backward compatibility

## Notes
Image Data Field now has Field label "No label" compared to others "Column name as label". Should we do anything about it?
Dynamic demo 2

### DIFF
--- a/js/default-configs/layouts-config.js
+++ b/js/default-configs/layouts-config.js
@@ -341,6 +341,14 @@ window.flListLayoutConfig = {
         column: 'Category'
       }
     ],
-    'showSummaryFieldsInDetailView': true
+    'showSummaryFieldsInDetailView': true,
+    'detail-fields': [
+      {
+        id: 'wWkqeYGN',
+        location: 'Image',
+        type: 'image',
+        column: 'Image'
+      }
+    ]
   }
 }


tonytlwu:
> Image Data Field now has Field label "No label" compared to others "Column name as label". Should we do anything about it?

@MaksymShokin It's OK. We can leave it as **No label**.

js/default-configs/layouts-config.js
```diff
@@ -341,6 +341,14 @@ window.flListLayoutConfig = {
         column: 'Category'
       }
     ],
-    'showSummaryFieldsInDetailView': true
+    'showSummaryFieldsInDetailView': true,
+    'detail-fields': [
```
tonytlwu: I'm not fully aware of how the property `showSummaryFieldsInDetailView` works, but if it does what it says, it seems to imply that all the fields in the`summary-fields` array will be shown in the detail view.

If that is the case, I can see that it already has an object for the image column:

```js
{
        id: 'gHlmJrlT',
        interfaceName: 'Image',
        location: 'Image',
        type: 'image',
        column: 'Image'
}
```

If that is the case, then why do we need to add a `detail-fields` array with just the Image field?

MaksymShokin: @tonytlwu Code in summary fields affects fields that are located in "Summary view field" column
Summary field

Code in detail fields affects "Detail view field"
Detail fields

PR #261, opened 2019-10-21
Deleted comment is now removed from UI right after deleting

MaksymShokin:
@tonytlwu @squallstar 

## Issue
Fliplet/fliplet-studio#5086

## Description
Corrected variable name. Updated confirm to fliplet.modal.confirm.

## Screenshots/screencasts
Delete demo

## Backward compatibility

### DIFF
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -2349,7 +2349,7 @@ DynamicList.prototype.replaceComment = function(guid, commentData, context) {
 DynamicList.prototype.deleteComment = function(id) {
   var _this = this;
   var entryId = _this.$container.find('.news-feed-list-item.open').data('entry-id') || _this.entryClicked;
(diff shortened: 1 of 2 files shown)


js/layout-javascript/news-feed-code.js
```diff
@@ -557,14 +557,17 @@ DynamicList.prototype.attachObservers = function() {
             {
               label: 'Delete',
               action: function (i) {
-                var options = {
+                Fliplet.Modal.confirm({
```
tonytlwu: We should only change to use `Fliplet.Modal` for Studio interface. This won't work in native apps.

squallstar: — as per note on developers documentation https://developers.fliplet.com/API/fliplet-core.html#modal

MaksymShokin: @tonytlwu Got it. Removed fliplet modal.

PR #264, opened 2019-10-23
Empty comments (or only spaces) will be filtered and not get rendered to UI

MaksymShokin:
@tonytlwu @squallstar 

## Issue
Fliplet/fliplet-studio#5084

## Description
Added filter to left out empty comments.

## Screenshots/screencasts
filter demo

## Backward compatibility

### DIFF
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -1988,13 +1988,16 @@ DynamicList.prototype.getEntryComments = function(options) {
         return instance.query({
           allowGrouping: true,
           where: {
-            content: content
+            content: content,
+            settings: {
+              text: { $regex: '[^\s]+' }
+            }
           }
         });
       })
       .then(function(entries) {
         var filteredEntries = _.filter(entries, function(entry) {
-          return entry.data.settings.text.trim();
+          return _.get(entry, 'data.settings.text', '').trim();
         });
         record.comments = filteredEntries;
         record.commentCount = filteredEntries.length;
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -1992,9 +1992,12 @@ DynamicList.prototype.getEntryComments = function(options) {
           }
         });
       })
-      .then(function(entries){
-        record.comments = entries;
-        record.commentCount = entries.length;
+      .then(function(entries) {
+        var filteredEntries = _.filter(entries, function(entry) {
+          return entry.data.settings.text.trim();
+        });
+        record.comments = filteredEntries;
+        record.commentCount = filteredEntries.length;
       });
   }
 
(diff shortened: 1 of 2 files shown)


js/layout-javascript/news-feed-code.js
```diff
@@ -1980,9 +1980,12 @@ DynamicList.prototype.getEntryComments = function(options) {
           }
         });
       })
-      .then(function(entries){
-        record.comments = entries;
-        record.commentCount = entries.length;
+      .then(function(entries) {
+        var filteredEntries = _.filter(entries, function(entry) {
+          return entry.data.settings.text.trim();
```
tonytlwu: We always like to be more careful with objects in case content is missing. The following allows any missing properties to return an empty string as a result, preventing errors.

```suggestion
          return _.get(entry, 'data.settings.text', '').trim();
```

Though, I think we can use queries to offload this to the backend. See my other comment.

tonytlwu: I couldn't comment on unmodified code. See a few lines up where a query object is provided. I'd change:

```js
instance.query({
  allowGrouping: true,
  where: {
    content: content
  }
})
```

...to...

```js
instance.query({
  allowGrouping: true,
  where: {
    content: content,
    settings: {
      text: { $regex: '[^\s]*' }
    }
  }
})
```

This is untested, but I imagine this will update the query so that it excludes empty text values right away. Use the developer network tool to confirm.

js/layout-javascript/simple-list-code.js
```diff
@@ -1934,9 +1934,12 @@ DynamicList.prototype.getEntryComments = function(options) {
           }
         });
       })
-      .then(function(entries){
-        record.comments = entries;
-        record.commentCount = entries.length;
+      .then(function(entries) {
+        var filteredEntries = _.filter(entries, function(entry) {
```
tonytlwu: See above

MaksymShokin: fixed

js/layout-javascript/news-feed-code.js
MaksymShokin: It works

js/layout-javascript/news-feed-code.js
```diff
@@ -1988,13 +1988,19 @@ DynamicList.prototype.getEntryComments = function(options) {
...
+              text: { $regex: '[^\s]+' }
+            }
           }
         });
       })
-      .then(function(entries){
-        record.comments = entries;
-        record.commentCount = entries.length;
+      .then(function(entries) {
+        var filteredEntries = _.filter(entries, function(entry) {
```
tonytlwu: Does the regex query work? If so, then we shouldn't need to filter the entries after they have been retrieved.

MaksymShokin: Yeah, it does work. Removed extra filter

PR #267, opened 2019-10-28
Edited comments now updated according to its ID

MaksymShokin:
@tonytlwu @squallstar 

## Issue
Fliplet/fliplet-studio#5085

## Description
The bug occurred because editing comments was updating the last comment. Now it updates comment by id.

## Screenshots/screencasts
editing comment demo

## Backward compatibility

### DIFF
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -2393,6 +2393,7 @@ DynamicList.prototype.saveComment = function(entryId, commentId, value) {
       return instance.update({
         settings: commentData.data.settings
       }, {
+        id: commentId,
         where: {
           content: content
         }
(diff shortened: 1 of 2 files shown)


js/layout-javascript/news-feed-code.js
```diff
@@ -2393,6 +2393,7 @@ DynamicList.prototype.saveComment = function(entryId, commentId, value) {
       return instance.update({
         settings: commentData.data.settings
       }, {
+        id: commentId,
```
tonytlwu: The `where` clause  beneath would be unnecessary if we pass the `id` property.

js/layout-javascript/simple-list-code.js
```diff
@@ -2347,6 +2347,7 @@ DynamicList.prototype.saveComment = function(entryId, commentId, value) {
       return instance.update({
         settings: commentData.data.settings
       }, {
+        id: commentId,
```
tonytlwu: The `where` clause  beneath would be unnecessary if we pass the `id` property.

js/layout-javascript/news-feed-code.js
MaksymShokin: fixed

js/layout-javascript/simple-list-code.js
MaksymShokin: fixed

PR #269, opened 2019-11-04
Aligned close icon for Firefox, Edge and Android 9 Chrome

YaroslavOvdii:
@tonytlwu @squallstar @sofiiakvasnevska 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/4773

## Description
Aligned close icon for Firefox, Edge and Android 9 Chrome.

## Screenshots/screencasts
Android 9 Chrome
Android9 Chrome

MS Edge
image

Firefox
Firefox

## Backward compatibility

### DIFF
--- a/css/build.css
+++ b/css/build.css
@@ -153,3 +153,25 @@
 .mode-interact [data-dynamic-lists-id] {
   min-height: 32px;
 }
+
+/* For MS Edge */
+@supports (-ms-ime-align:auto) {
+  .fa-times-thin {
+    padding-top: 1px;
+    padding-bottom: 3px;
+  }
+}
+
+/* For Firefox */
+@-moz-document url-prefix() {
+  .fa-times-thin {
+    padding-top: 2px;
+    padding-bottom: 5px;
+  }
+}
+
+/* For Android Chrome browser*/
+.android.no-native .fa-times-thin {
+  padding-top: 2px;
+  padding-bottom: 2px;
+}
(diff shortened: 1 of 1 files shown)


tonytlwu:
@YaroslavOvdii 9 PRs have been opened to attempt to resolve this issue [since 6th August](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/220). Can you explain why this was so challenging and how you are 100% certain there won't be a 10th PR? cc @sofiiakvasnevska

YaroslavOvdii:
@tonytlwu 
Challenge, in this case, is that we are using the Unicode icon instead of the SVG icon which looks differently in browsers and have different height and width.
We are trying to compensate this by adding paddings to these icons.
9 PR's were made because of my mistake, in this case, I wasn't looking pixel-perfect at first and wasn't tested it on all available platforms. Now I tested this PR on all available for me platforms such as:
Windows: 
1. Chrome
2. MS Edge
3. MS IE11
4. Firefox

macOS(virtual machine): 
1. Safari
2. Chrome
3. Firefox

Android 9 device: 
1. Chrome
2. Native

I can't say that it would be 100% last PR for this task. I tried to take account of everything I could imagine but there may be something I didn't think about.

Before sending this message I checked one more time on real macOS and found an issue with positioning on firefox. 
I've fixed it in the next commit.

tonytlwu:
@YaroslavOvdii You can raise and mention the challenge next time. We would've recommended switching to SVG icon if it's easier to center with less code. I fear that with a case-by-case padding adjustment, it'll screw up as soon as any rendering engine or OS upgrade changes how unicode's rendered.

PR #274, opened 2019-11-07
Simple list description in IE11 will look like in other browsers

YaroslavOvdii:
@tonytlwu @squallstar @sofiiakvasnevska 
## Issue

## Description
Simple list description in IE11 will look like in other browsers

## Screenshots/screencasts
issue 5030

## Backward compatibility

### DIFF
--- a/css/layout-css/simple-list-style.upload.css
+++ b/css/layout-css/simple-list-style.upload.css
@@ -556,6 +556,7 @@
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
   overflow: hidden;
+  height: 4.4em;
 }
 
 .simple-list-item .list-item-title {


tonytlwu:
The description says:

> Simple list description in IE11 will look like in other browsers.

This doesn't look right. There's all the extra undesirable padding and white space in the new screenshot. It looks like the simple list description will look **different** in IE 11 from other browsers.

image

For reference, this is how it looks in Chrome and there's none of the extra padding.

image

The reported issues mentions that there's no clipping and no ellipsis. The first option should be to add a max height and ensure the ellipsis are displayed if it goes beyond 3 lines.

YaroslavOvdii:
@tonytlwu 
I've tried to reproduce this issue, but with the updated repository from 08.11, I can't see it.

image

About ellipsis, IE11 doesn't [support ](https://caniuse.com/#search=display%3A%20-webkit-box) `display: -webkit-box;` and if we try to add ellipsis with after element we will have issue with line width. For example:
image

As you may see all entries has defferent width and when we try to add ellipsis we wan't be able to gess where text line will end. 

Therefore may we not to add ellipsis for IE11?
Or look some workout for the IE11?

YaroslavOvdii:
@tonytlwu created new [PR ](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/283) with [this solution](https://github.com/Fliplet/fliplet-studio/issues/5030#issuecomment-553406237)

PR #285, opened 2019-11-28
Show normaly clipped news feed layout to a user

YaroslavOvdii:
@sofiiakvasnevska 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/5373

## Description
Show normally clipped news feed layout to a user

## Screenshots/screencasts
image

## Backward compatibility

## Notes
It seems that `display: -webkit-box` doesn't work right there is more than one child in it. The only way I've found to fix this it was by adding the `max-height` attribute.
I didn't add it threw the javascript because I wasn't able to found this kind of solution in the library we are using.
I've added this solution as well to the simple list because when a user will try to do the same description in the simple list he will counter the same issue.

### DIFF
--- a/css/layout-css/news-feed-style.upload.css
+++ b/css/layout-css/news-feed-style.upload.css
@@ -775,6 +775,11 @@
   line-height: 1.5;
 }
 
+.new-news-feed-list-container .news-feed-list-item:not(.open) .slide-over .news-feed-item-inner-content .news-feed-item-description,
+.new-news-feed-list-container .news-feed-list-item:not(.open) .slide-over .news-feed-item-inner-content .news-feed-item-title {
+  max-height: 4.4em;	
+}
+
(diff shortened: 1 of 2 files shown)


tonytlwu:
I don't like having to set a `max-height`. The benefit of having the `data-line-clamp="3"` property is that users can potentially change the value or remove the clamping if they wish. I don't want a `max-height` to be associated with it, especially if font sizes and so on could be changed.

If `display: -webkit-box` is the cause of the problem, we have 2 options:

1. Look at why we use `display: -webkit-box` and achieve the same results another way; or
1. Resolve the ellipsis issue in a different way

tonytlwu:
Changes are still needed from my previous comment. Please suggest a recommended solution.

Also, this needs conflicts resolved before a review.

YaroslavOvdii:
@tonytlwu 
I'm still looking for a solution which will suit this issue.

If you don't mind may I close this PR and create a new one with the preferable solution?

PR #288, opened 2019-12-03
Returned CSS code to support fa-times-thin class for close icon

YaroslavOvdii:
@sofiiakvasnevska 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/5347

## Description
Returned CSS code to support fa-times-thin class for the close icon according to this [comment](https://github.com/Fliplet/fliplet-studio/issues/5347#issuecomment-558279446).
## Backward compatibility

## Notes
I wasn't able to test it properly because of cant replicate the issue.

### DIFF
--- a/css/build.css
+++ b/css/build.css
@@ -122,19 +122,19 @@
   to use UNICODE icons instead of SVG icons.
 */
 
-.close-icon .fa-times-thin {
+[data-widget-package="com.fliplet.dynamic-lists"] .close-icon .fa-times-thin {
   padding-top: 2px;
   padding-bottom: 4px;
   font-size: 0.8em !important;
 }
 
-.close-icon .fa-times-thin::before {
+[data-widget-package="com.fliplet.dynamic-lists"] .close-icon .fa-times-thin::before {
   content: '\002573';
 }
 
 /* IE11 layout */
 @media screen and (-ms-high-contrast: none) {
-  .close-icon .fa-times-thin {
+  [data-widget-package="com.fliplet.dynamic-lists"] .close-icon .fa-times-thin {
     padding-top: 0px;
     padding-bottom: 5px;
   }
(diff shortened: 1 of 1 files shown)


css/build.css
```diff
@@ -115,4 +115,72 @@
 
 .close-icon {
   width: 40%;
+}
+
+.fa-times-thin {
```
tonytlwu: Add a comment above this section to explain why this was necessary. Otherwise, future developers may find this code unnecessary and delete them.

YaroslavOvdii: @tonytlwu Added a comment above `.fa-times-thin` section start.

css/build.css
```diff
@@ -115,4 +115,77 @@
 
 .close-icon {
   width: 40%;
+}
+
+/*
+  This part of the code needed for support .fa-times-thin class in users who manage to update the LFD component
```
tonytlwu: Use this copy. Split into multiple lines as necessary.

```suggestion
  This part of the code is needed for supporting .fa-times-thin class in users who updated to use UNICODE icons instead of SVG icons.
```

YaroslavOvdii: @tonytlwu Updated the comment

css/build.css
```diff
@@ -115,4 +115,77 @@
...
 .close-icon {
   width: 40%;
+}
+
+/*
+  This part of the code is needed for supporting .fa-times-thin class in users who updated
+  to use UNICODE icons instead of SVG icons.
+*/
+
+.fa-times-thin {
```
tonytlwu: I think it's best to use a selector alongside this, e.g. `.close-icon .fa-times-thin`. We don't want all of `.fa-times-thin` anywhere in the app to be affected -- only if it's used in the LFD within the expected detail view overlay.

YaroslavOvdii: @tonytlwu 
Added more specified CSS selector

css/build.css
```diff
@@ -122,19 +122,19 @@
   to use UNICODE icons instead of SVG icons.
 */
 
-.fa-times-thin {
+.close-icon .fa-times-thin {
```
tonytlwu: I think we can be more specific than this.

`[data-widget-package="com.fliplet.dynamic-lists"]` would be a pretty good selector to add so that it only affects `.fa-times-thin` within LFDs.

YaroslavOvdii: @tonytlwu  Agree.
Added data attribute to the selector for a more specified selector.

PR #291, opened 2019-12-09
Fixed saving data view settings

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5312

## Description
When user deletes field, id is removed from the array.

## Backward compatibility

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -528,6 +528,10 @@ var DynamicLists = (function() {
           var fieldId = $(this).parents('.rTableRow').data('id');
           var $row = $(this).parents('.rTableRow');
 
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -528,6 +528,10 @@ var DynamicLists = (function() {
           var fieldId = $(this).parents('.rTableRow').data('id');
           var $row = $(this).parents('.rTableRow');
 
+          selectedFieldId = _.filter(selectedFieldId, function(item) {
```
tonytlwu: Use `_.remove()` here

MaksymShokin: fixed

PR #292, opened 2019-12-11
Remove unnecessary margins

hcarneiro:
Removed these unnecessary margins, now that we have better control with the Appearance settings.

Annotation 2019-12-11 145622


hcarneiro:
@benws1 See Tony's comment.

benws1:
@tonytlwu @hcarneiro - this change should be queued somewhere for when a new version is created.

PR #293, opened 2019-12-12
Show ellipsed description in all browsers.

YaroslavOvdii:
@sofiiakvasnevska 
## Issue

## Description
Show ellipsed description in all browsers.

## Screenshots/screencasts
Simple list IE11
simple IE11

News feeds Chrome
Chrome

News feeds Edge
Edge

News feeds Firefox
Firefox

News feeds IE11
IE11

Simple list Chrome
simple chrome

Simple list Edge
simple edge

Simple list Firefox
simple firefox

How it looks.
view

## Backward compatibility


js/layout-javascript/news-feed-code.js
```diff
@@ -1479,6 +1479,7 @@ DynamicList.prototype.renderLoopHTML = function (iterateeCb) {
         requestAnimationFrame(render);
       } else {
         _this.$container.find('.new-news-feed-list-container').removeClass('loading').addClass('ready');
+        ellipsizeTextBox('.news-feed-item-description', 3);
```
tonytlwu: Is this the same library as before or a different one?

Also, can we use the same `data-line-clamp` attribute as we did with the previous solution so that users can configure this via HTML template?

YaroslavOvdii: @tonytlwu 
No, this is not the same library. I'll add a simple function that reduces the inner HTML of the element and add a dots at the end. 

> Also, can we use the same data-line-clamp attribute as we did with the previous solution so that users can configure this via HTML template?

Yes, we can use any CSS selector. I'll add it right away.

YaroslavOvdii: @tonytlwu 
Added `data-line-clamp` to the template.

vendor/ellipsis.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Method to clip the content of the element
+ * Doesn't work with inline elements
+ * 
+ * @param {string} selector - a CSS selector for the element that needed to be clipped 
+ * @param {number} lines - number of lines that need to left
+ */
+function ellipsizeTextBox(selector, lines) {
+   var lineHeight = 1.2; // Default line height in em
```
tonytlwu: Best to calculate real line height using `computeStyle()`.

vendor/ellipsis.js
```diff
@@ -0,0 +1,39 @@
...
+
+   if (el.length === 0) {
+      console.error('There is no elements on the page with this "' + selector + '" selector. Please check the selector.');
+      return
+   }
+
+   for (var i = 0; i < el.length; i++) {
+      var isInline = _.indexOf(inlineTags, el[i].tagName.toLowerCase());
+
+      if (isInline > -1) {
```
tonytlwu: Perhaps we can calculate whether the element is inline via `computeStyle()`, because users could be using custom CSS to change the `display` property of an element.

vendor/ellipsis.js
```diff
@@ -0,0 +1,39 @@
...
+         console.error('Please, do not use ellipsizeTextBox function on inline elements it may cause an infinite loop and block your app.',
+            'You have used it on <' + el[i].tagName.toLowerCase() + '> tag which was found by this "' + selector + '" selector');
+         break;
+      }
+
+      el[i].style.height = elHeight + 'em'; // Set element height according to the number of lines we want to left
```
tonytlwu: What if the line height was set using absolute `px` numbers?

js/layout-javascript/news-feed-code.js
tonytlwu: I see. I've reviewed the library as well. I see a few problems with it.

Which was the library we previously used?

YaroslavOvdii: @tonytlwu 
We were using [this](https://github.com/josephschmitt/Clamp.js/) library previously.

tonytlwu: @YaroslavOvdii What other library have you tried?

I found these two, and I wonder if they would be helpful.

https://github.com/nzambello/ellipsed
https://github.com/makandra/superclamp

YaroslavOvdii: @tonytlwu If I may ask. Should we make one more feature with line clamp?
I probably wasn't looking into this issue from the perspective of improvement.

As far as I tested both libs you suggested, a better option for the further improvement would be a [superclamp](https://github.com/makandra/superclamp) lib.

And [this](https://github.com/nzambello/ellipsed) is similar to our current solution. 

But if we want just fix this issue. I would prefer to use a simpler solution which we could easier adjust to our needs. 

I agree that our solution may look raw but it can change easily to our needs and if you give us requirements of how the function should work we can make it great.

About the new feature, we could, for example, allow users to implement clamp-on text elements on any widget with an option.

tonytlwu: > About the new feature, we could, for example, allow users to implement clamp-on text elements on any widget with an option.

This is an interesting approach. I can see something like this being useful. In our usual fashion, however, we'd implement this to be specific to the LFD first and roll it out more globally if people feel the need to control line clamping across the app.

I would therefore suggest using [superclamp](https://github.com/makandra/superclamp) to implement this for the LFD, using `data-line-clamp` attributes on the containers to control the maximum number of lines to show.

Although a self-rolled solution might also work, based on my comments on your last round of implementation I'm not confident in the quality. I'd therefore be more confident in using a library that's been used more and for longer.

YaroslavOvdii: @tonytlwu I'll use the superclamp then. I'll add it by my next commit.

js/layout-javascript/simple-list-code.js
```diff
@@ -1205,6 +1205,20 @@ DynamicList.prototype.renderLoopHTML = function (iterateeCb) {
         requestAnimationFrame(render);
       } else{
         _this.$container.find('.simple-list-container').removeClass('loading').addClass('ready');
+
+        // SuperClamp works only when elemet has a fixed max-height or height
+        // To do so we need to take computed line-height and multiple it on data-line-clamp value
+        setClampHeight('[data-line-clamp]')
```
tonytlwu: Where is the `setClampHeight` function defined?

YaroslavOvdii: @tonytlwu it is defined in the `build.js` file. I moved there to prevent code duplicate because we need it in both layouts.

tonytlwu: I don't see it in the `build.js` file. Did you commit the changes?

Also, you should use the `utils.js` file for commonly used functions in the LFD.

YaroslavOvdii: @tonytlwu I've tried to add these functions to the `utils.js` but when I test it at the line that you highlighted function doesn't declare yet.
Is there any specific to add functions to this file?

I've added my functions into the body of the `Fliplet.Registry.set('dynamicListUtils', (function () {`

tonytlwu: @YaroslavOvdii I think I know why it doesn't work. but I can't comment on it unless you've committed your code.

YaroslavOvdii: @tonytlwu I think I figure it out why it doesn't work earlier.
Added new commit.

js/layout-javascript/news-feed-code.js
```diff
@@ -1479,6 +1479,20 @@ DynamicList.prototype.renderLoopHTML = function (iterateeCb) {
         requestAnimationFrame(render);
       } else {
         _this.$container.find('.new-news-feed-list-container').removeClass('loading').addClass('ready');
+
+        // SuperClamp works only when elemet has a fixed max-height or height
+        // To do so we need to take computed line-height and multiple it on data-line-clamp value
+        _this.Utils.Page.setClampHeight('[data-line-clamp]')
```
tonytlwu: All of this

```js
        _this.Utils.Page.setClampHeight('[data-line-clamp]')
          .then(function (elements) {
            Superclamp.register(elements);
            return Promise.resolve(elements);
          })
          .then(function (elements) {
            _this.Utils.Page.removeClampHeight(elements);
          })
          .catch(function (selector) {
            console.error('Fail to clamp line in the selector ' + selector);
          });
```

...seems to be the same. Why not just create a `_this.Utils.Page.setLineClamps()` function that returns a promise?

js/utils.js
```diff
@@ -1245,7 +1294,8 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     },
     Page: {
       updateSearchContext: updateSearchContext,
-      updateFilterControlsContext: updateFilterControlsContext
+      updateFilterControlsContext: updateFilterControlsContext,
+      setLineClamps: setLineClamps,
```
tonytlwu:
```suggestion
      setLineClamps: setLineClamps
```

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
   };
   var computedFieldClashes = [];
 
+  function setLineClamps () {
+    setClampHeight()
```
tonytlwu:
```suggestion
    return setClampHeight()
```

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+      .then(function (elements) {
+        Superclamp.register(elements);
+        return Promise.resolve(elements);
```
tonytlwu:
```suggestion
        return elements;
```

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+      })
+      .then(function (elements) {
+        removeClampHeight(elements);
```
tonytlwu: Why not just add this to the previous `.then()`? There doesn't appear to be anything that needs promise chaining here.

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+      resolve(elements);
+    });
+  }
+
+  // To avoid large distance between block we remove height at the element
+  function removeClampHeight (elements) {
+    for (var i = 0; i < elements.length; i++) {
+      var element = elements[i];
+
+      element.removeAttribute("style");
```
tonytlwu: What if the element contains other styles? It should only be the height that gets removed.

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+      .catch(function (selector) {
+        console.error('Fail to clamp line in the selector ' + selector);
+      });
+  }
+
+  function setClampHeight (selector) {
+    selector = selector || '[data-line-clamp]';
+
+    return new Promise(function (resolve, reject) {
+      var elements = document.querySelectorAll(selector);
```
tonytlwu: I suggest using jQuery as it supports a wider range of selectors.

```suggestion
      var $elements = $(selector);
```

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+      
+      if (!elements.length) {
+        reject(selector);
+      }
+
+      for (var i = 0; i < elements.length; i++) {
```
tonytlwu: If you use jQuery, you can use `$elements.each()` here.

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+        var element = elements[i];
+        var lineHeightValue = window.getComputedStyle(element)
```
tonytlwu: What does the `.match(/[a-zA-Z]+|[0-9]+(?:\.[0-9]+|)/g)` part do?

Does it always return a number in pixels? If not, then what are the possible values and how are we dealing with each of the possible value types/units?

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
   };
   var computedFieldClashes = [];
 
+  function setLineClamps () {
```
tonytlwu: Since setClampHeight supports a `selector` parameter...

```suggestion
  function setLineClamps (selector) {
```

js/utils.js
tonytlwu:
```suggestion
    setClampHeight(selector)
```

js/utils.js
```diff
@@ -17,6 +17,55 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+      })
+      .then(function (elements) {
+        removeClampHeight(elements);
+      })
+      .catch(function (selector) {
+        console.error('Fail to clamp line in the selector ' + selector);
+      });
+  }
+
+  function setClampHeight (selector) {
```
tonytlwu: Can you explain again (after so many rounds of changes), why we need to do this, now that we're using Superclamp?

js/utils.js
YaroslavOvdii: @tonytlwu 

> Does it always return a number in pixels?

I've tested it with different values as `em` or `rem` and it always returns the value in pixels.

> What does the .match(/[a-zA-Z]+|[0-9]+(?:\.[0-9]+|)/g) part do?

This part splits the code to array with value and unit.

js/utils.js
YaroslavOvdii: @tonytlwu 
As SuperClamp says in the [readme file](https://github.com/makandra/superclamp#usage)

> You want to truncate text to fit inside an element, so define your width/height restrictions on the element, e.g. via CSS. Superclamp will respect an element's padding.

That's why we need to set a fixed height value to the element.

js/utils.js
```diff
@@ -17,6 +17,50 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+    selector = selector || '[data-line-clamp]';
+
+    return new Promise(function (resolve, reject) {
+      var $elements = $(selector);
+      
+      if (!$elements.length) {
+        reject(selector);
```
tonytlwu: I think this needs a `return` after this to stop the execution.

```suggestion
        reject(selector);
```

js/utils.js
```diff
@@ -17,6 +17,50 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+      });
+
+      resolve($elements);
+    });
+  }
+
+  // To avoid large distance between block we remove height at the element
+  function removeClampHeight ($elements) {
+    $elements.each(function() {
+      $(this).css('height', 'auto');
```
tonytlwu: This is the better way to remove the `height` property instead of setting it to `auto`. That's because if there is CSS setting the height to, e.g. `3em`, then removing the clamp height should reset it to `3em` by removing the inline CSS, instead of changing `3em` to `auto`.

```suggestion
      $(this).css('height', '');
```

PR #296, opened 2019-12-27
Fixed collapse target error

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5155

## Description
Added check if collapseTarget has offset method.

## Backward compatibility

### DIFF
--- a/js/layout-javascript/small-card-code.js
+++ b/js/layout-javascript/small-card-code.js
@@ -1744,7 +1744,7 @@ DynamicList.prototype.collapseElement = function(elementToCollapse) {
 
   var collapseTarget = elementToCollapse.parent();
 
-  if (!collapseTarget || !collapseTarget.offset()) {
+  if (!collapseTarget.length) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


js/layout-javascript/small-card-code.js
```diff
@@ -1743,6 +1743,11 @@ DynamicList.prototype.collapseElement = function(elementToCollapse) {
   }
 
   var collapseTarget = elementToCollapse.parent();
+
+  if (!collapseTarget || !collapseTarget.offset()) {
```
tonytlwu: If `elementToCollapse` was able to `.parent()`, it means that `elementToCollapse` is a jQuery object.

This means `collapseTarget` will always be a jQuery object too and truthy. You should check for `collapseTarget.length` instead. If you test it on jquery.com you'll find that `$('asdkhasdjsad').offset()` returns `undefined`, which means the problem here most likely comes from `collapseTarget` is an empty array.

```suggestion
  if (!collapseTarget.length) {
```

MaksymShokin: Got it! Fixed

PR #300, opened 2020-01-13
Now images with similar titles display correctly

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5447

## Description
Before the issue occurred because of indexOf check. 
Now we will find the last dot in the image name, remove ext  and then compare names.

## Screenshots/screencasts
photos-lfd

## Backward compatibility

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -792,7 +792,10 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
             return;
           }
 
-          if (entryData[column] && file.name.indexOf(entryData[column]) !== -1) {
+          var lastDotInNameIndex = file.name.lastIndexOf('.');
+          var nameWithoutExt = file.name.substring(0, lastDotInNameIndex);
+
+          if (entryData[column] && (file.name === entryData[column] || nameWithoutExt === entryData[column])) {
             // File found
             entryData[column] = file.url;
             return true;
--- a/js/utils.js
+++ b/js/utils.js
@@ -792,10 +792,10 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
             return;
           }
 
-          var lastDotInNameIndex = file.name.lastIndexOf('.');
-          var nameWithoutExt = file.name.substring(0, lastDotInNameIndex);
+          // remove file extension 
+          var fileName = file.name.match(/(.+?)(?:\.[^\.]*$|$)/)[1];
 
-          if (entryData[column] && (file.name === entryData[column] || nameWithoutExt === entryData[column])) {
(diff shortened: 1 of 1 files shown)


js/utils.js
```diff
@@ -792,7 +792,10 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
             return;
           }
 
-          if (entryData[column] && file.name.indexOf(entryData[column]) !== -1) {
+          var lastDotInNameIndex = file.name.lastIndexOf('.');
+          var nameWithoutExt = file.name.substring(0, lastDotInNameIndex);
```
tonytlwu: What happens when `lastDotInNameIndex` is `-1`?

image

This looks like it won't work when the file name doesn't contain a `.`.

We should also be able to extract this with a regexp.

MaksymShokin: @tonytlwu In line 798 we first compare the file name without extension and then with removed extension. I tested this and it works fine.

tonytlwu: I see. Then this is confusing because when a `.` isn't found, the line `nameWithoutExt = file.name.substring(0, lastDotInNameIndex)` not only doesn't make sense, but is also unnecessary.

There should be regex that's available to get the file name, e.g.

```js
var fileName = file.name.match(/(.+?)(?:\.[^\.]*$|$)/)[1];
```

If this doesn't work, research and find the appropriate solution. Otherwise, this logic feels odd as I described above.

MaksymShokin: updated to use regexp

js/utils.js
```diff
@@ -792,7 +792,10 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
             return;
           }
 
-          if (entryData[column] && file.name.indexOf(entryData[column]) !== -1) {
+          // remove file extension 
+          var fileName = file.name.match(/(.+?)(?:\.[^\.]*$|$)/)[1];
+
+          if (entryData[column] && (file.name === entryData[column] || fileName === entryData[column])) {
```
tonytlwu: image

```suggestion
          if (entryData[column] && fileName === entryData[column]) {
```

MaksymShokin: In that case, it would not fully work, because the user has the possibility to write entryData[column] with extension. 
For example, dog.jpeg , in that case file.name after regexp would be simply dog.

MaksymShokin: We use regexp on file names, which always do have the extension. But user can enter both dog and dog.jpeg

tonytlwu: Ah yes, you're right 👍

PR #303, opened 2020-01-17
Speed up interface load time if a layout is already selected

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/5131

A `setupLayoutSelector()` function is created to set up the layout selector, which only does anything if a layout isn't already selected. This is previously being executed (and layout selector added to the DOM) even if the UI never gets used. This loads 28.4MBs of GIF into the browser and slows down the overall time to render.

Locally, this has reduced the time to render from ~8.5s to ~4.5s.

On staging, the time to render is currently ~3s.
On production, the time to render is currently ~7s.

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -82,10 +82,9 @@ var DynamicLists = (function() {
       social: {},
       advancedSettings: {}
     }, configuration);
-    _this.widgetId = configuration.id;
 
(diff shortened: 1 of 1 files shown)


tonytlwu:
**Update** The production render time has reduced from ~7s to ~3s.

PR #306, opened 2020-01-27
When users will return form booked to all entries there text won't jump in the process

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5030#issuecomment-575585397

## Description
When users will return form booked to all entries there text won't jump in the process.

## Screenshots/screencasts
5030-glitch

## Backward compatibility

## Notes
We need to set default height so that clamped component do not jump when redraw happens.


css/build.css
```diff
@@ -188,4 +188,9 @@
 .android.no-native [data-widget-package="com.fliplet.dynamic-lists"] .close-icon .fa-times-thin {
   padding-top: 2px;
   padding-bottom: 2px;
+}
+
+[data-line-clamp].clamp {
```
tonytlwu: If we use the below, then we can cover some common scenarios. I've set the default to 140px so that the default would be 10 lines, before possibly expanding to more.

And, we don't need the `.clamp` part, right?

```suggestion
[data-line-clamp] {
  height: 140px;
}

[data-line-clamp="1"] {
  height: 14px;
}

[data-line-clamp="2"] {
  height: 28px;
}

[data-line-clamp="3"] {
  height: 42px;
}

[data-line-clam="4"] {
  height: 56px;
}

[data-line-clamp="5"] {
  height: 70px;
}

[data-line-clamp="6"] {
  height: 84px;
}

[data-line-clamp="7"] {
  height: 98px;
}

[data-line-clamp="8"] {
  height: 112px;
}

[data-line-clamp="9"] {
  height: 126px;
}

[data-line-clamp="10"] {
  height: 140px;
}
```

YaroslavOvdii: @tonytlwu `.clamp` part is needed in this scenario because we are removing this class from the clamped element and it would have it own height not we have set by css. In another case we might have gasp between elements

tonytlwu: Instead of having the extra `.clamp` to start with and the removing it, we could remove the `[data-line-clamp]` attribute after an inline height has been set. Will that work?

The current solution has the elegance of telling developers, "you just need to add `[data-line-clamp="x"]`" and ideally I don't want to change that to, "you just need to add `[data-line-clamp="x"]` and also `.clamp`".

YaroslavOvdii: @tonytlwu Yes, if we remove `[data-line-clamp]` it will work correctly.
Should I made a new commit with this solution?

tonytlwu: Yes please, and please also add the CSS default for the various line clamp values as I have recommended.

YaroslavOvdii: @tonytlwu 
Added these changes, also add the `overflow: hidden` the main `[data-line-clamp]`.

PR #308, opened 2020-01-30
Adds method for customising primary key and entry identifiers for social features

tonytlwu:
Also follows https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/299 to adjust some agenda layout spacing.

To define a custom primary key:

```js
Fliplet.Hooks.on('flListDataBeforeGetData', function (options) {
  options.config.dataPrimaryKey = 'Session ID';
});
```

To create a custom primary key based on various data:

```js
Fliplet.Hooks.on('flListDataBeforeGetData', function (options) {
  options.config.dataPrimaryKey = function (data) {
    return _.get(data, 'record.data.Title') + '-' + _.get(data, 'record.data.Date');
  };
});
```

### DIFF
--- a/css/layout-css/agenda-style.upload.css
+++ b/css/layout-css/agenda-style.upload.css
@@ -849,16 +849,29 @@
   padding-bottom: calc(65px + env(safe-area-inset-bottom));
 }
 
-[data-has-notch] .new-agenda-list-container.has-add-entry .agenda-cards-wrapper .agenda-list-day-holder .agenda-list-card-holder {
-  padding-bottom: calc(65px + 34px);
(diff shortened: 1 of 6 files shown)


squallstar:
@tonytlwu why does the primary key attribute needs the `data` prefix? Could there be a `primaryKey` applying to a different entity?

tonytlwu:
@squallstar I don't have any scenario for it. However, the `config` object describes the configuration for the LFD instance, which is feature rich. Not that I'd know why yet, but primary keys could possibly apply to `comments`, `users` etc.

PR #312, opened 2020-02-04
Adjusts layout so filters and results are scrollable together

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/5714


css/layout-css/simple-list-style.upload.css
```diff
@@ -409,6 +409,8 @@
   width: 100%;
   display: none;
   align-items: center;
+  color: #aaaaaa;
```
squallstar: `#aaa` would have been cleaner

PR #325, opened 2020-02-24
Support for segmentation user bookmarks

squallstar:
This PR adds support for
- creating a user's bookmarks view definition when creating the bookmarks data source (bundled by default)
- all internal features to use the above newly defined view
- adds new reference type for views

Existing usage will be unaffected as the bookmarks data sources won't have the view definition, resulting in simply skipping these inputs from having any effect.

---

requires https://github.com/Fliplet/fliplet-api/pull/3866

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -2541,8 +2541,14 @@ var DynamicLists = (function() {
                 filter: {
                   user: {
                     $or: [
+                      // Guest users
                       { uuid: 'session.uuid' },
-                      { email: 'session.email' }
+
+                      // Logged in users
+                      { email: 'session.email' },
+
+                      // Legacy support just in case
+                      { dataSourceEntryId: 'session.id' }
                     ]
                   }
                 }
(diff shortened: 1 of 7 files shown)


js/interface.js
```diff
@@ -2528,13 +2528,42 @@ var DynamicLists = (function() {
...
+                filter: {
+                  user: {
+                    $or: [
+                      // Guest users
+                      { uuid: 'session.uuid' },
+
+                      // Logged in users
+                      { email: 'session.email' },
+
+                      // Legacy support just in case
```
squallstar: this is made so that for LMA we can update the view definition manually and have existing bookmarks working.

js/interface.js
```diff
@@ -2528,13 +2528,42 @@ var DynamicLists = (function() {
...
+          }
         }).then(function (dataSource) {
           _this.config.bookmarkDataSourceId = dataSource.id;
+          _this.config.dataSourceViews = _this.config.dataSourceViews || [];
+
+          _this.config.dataSourceViews.push({
+            id: dataSource.id,
+            name: 'userBookmarks',
+            bundle: true
+          })
```
tonytlwu: Let's run a `_.uniq` through `_this.config.dataSourceViews` to make sure we don't have any duplicated views of the same data source. The mechanism for managing bookmark/like/comment data sources don't exist at the moment, but I'd play it safe just in case?

squallstar: @tonytlwu done, check latest commit.

PR #327, opened 2020-02-26
Clamped text with ellipsis in IE11 and other browsers

YaroslavOvdii:
@sofiiakvasnevska 

## Issue

## Description
Clamped text with ellipsis in IE11 and other browsers

## Screenshots/screencasts
image

## Backward compatibility


css/layout-css/news-feed-style.upload.css
```diff
@@ -1544,3 +1536,11 @@
     display: none;
   }
 }
+
+.new-news-feed-list-container.loading .news-feed-list-wrapper {
```
tonytlwu: If I remember correctly, when most (if not all) LFD layouts are loading, there's a loading state for which it says "Loading..." on the screen. Is this the state that you're applying `opacity: 0` to?

YaroslavOvdii: @tonytlwu Yes, I'm using the `.loading` class to display for a user "Loading" while SuperClamp library is clamping the lines so that the user wouldn't see jumping lines.

tonytlwu: @YaroslavOvdii The "Loading..." state is important to be visible as the LFD instance could be loading content from the data source API, which takes time. We wouldn't want this state to be invisible.

The `renderLoopHTML()` function, in these layouts, would give you the logic to show what happens right before and after content is rendered. I recommend using the beginning and end of this process to apply opacity 0 and 1 to the list, instead of hiding the whole thing including the loading state.

YaroslavOvdii: @tonytlwu No, the loading state isn't hiding. We hide only the list with opacity but a user will see the "Loading..." state.

PR #333, opened 2020-03-17
Returned CSS code for old templates of the LFD so text would clamp properly

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6062

## Description
Returned CSS code for old templates of the LFD so text would clamp properly

## Backward compatibility

### DIFF
--- a/css/layout-css/news-feed-style.upload.css
+++ b/css/layout-css/news-feed-style.upload.css
@@ -779,6 +779,14 @@
   left: 0;
 }
 
+.new-news-feed-list-container .news-feed-list-item:not(.open) .slide-over .news-feed-item-inner-content .news-feed-item-description {	
+  overflow: hidden;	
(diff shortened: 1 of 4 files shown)


tonytlwu:
@YaroslavOvdii Can we enhance this even further so that the JS solution is only used for IE 11?

YaroslavOvdii:
@YaroslavOvdii Yes it is possible, we could use the second option from this [suggestion ](https://github.com/Fliplet/fliplet-studio/issues/6062#issuecomment-599640798) and use Modernizr to find out what browser user is using.
Should I try to implement this suggestion?

tonytlwu:
@YaroslavOvdii Yes, let's do that.

YaroslavOvdii:
@tonytlwu Added code that we talk about above.

PR #334, opened 2020-03-18
Speeding up LFD load time

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/5817

- On initial data load, instead of preparing all summary view and detail view data, only add filter properties for adding filters onto the screen.
- When rendering the list view, only process records for the summary view data (skipping the detail view data processing).
- When opening an entry to show its details, add detail view data for the single record.

## Affected user flow

- Initial load
- Render time (see https://github.com/Fliplet/fliplet-studio/issues/5817#issuecomment-592224853 for detailed benchmarks)
- Opening an entry
- Filtering, search and bookmark filter
- Social features (bookmark, like, comment)
- Opening an entry on load (using the `dynamicListOpenId` parameter)
- Determining who has access to insert, edit and delete entries

### DIFF
--- a/js/layout-javascript/agenda-code.js
+++ b/js/layout-javascript/agenda-code.js
@@ -815,7 +815,7 @@ DynamicList.prototype.initialize = function() {
       });
     })
     .then(function(records) {
-      _this.listItems = records;
+      _this.listItems = _this.getPermissions(records);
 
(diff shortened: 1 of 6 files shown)


squallstar:
@tonytlwu can you outline the performance gains before and after the patch?

tonytlwu:
@squallstar See https://github.com/Fliplet/fliplet-studio/issues/5817#issuecomment-592224853

PR #340, opened 2020-03-24
Adds sortable functionality to detail view fields

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/6126

## Screenshot

image

### DIFF
--- a/css/interface.css
+++ b/css/interface.css
@@ -425,7 +425,8 @@ header .btn,
 }
 
 .panel-title .fa {
-  cursor: pointer;
+  cursor: -webkit-grab;
+  cursor: -moz-grab;
 }
 
 .panel-chevron {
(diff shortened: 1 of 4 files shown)


css/interface.css
```diff
@@ -425,7 +425,8 @@ header .btn,
 }
 
 .panel-title .fa {
-  cursor: pointer;
+  cursor: -webkit-grab;
+  cursor: -moz-grab;
```
squallstar: no `grab` ?

hcarneiro: Same on line `920`.

tonytlwu: Apparently not. I'm following the existing patterns, which seem to be working quite well.

hcarneiro: Weird, but ok

PR #346, opened 2020-03-25
Add customisable session ID queries to agenda links

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/6130

The `dataPrimaryKey` configuration was built to be used for bookmarking. This PR extends the feature to be used for polls, surveys and questions too. When going to the poll, survey and question screens, the `sessionId` query will be added to either send the data source entry ID or the primary key according to the `dataPrimaryKey` configuration.


tonytlwu:
@squallstar I'm still running tests. I'll see what's efficiently possible as this was an unplanned enhancement to avoid technical debt on the app templates.

PR #364, opened 2020-04-15
Agenda search wrapper now looks correctly on IE on tablet and desktop

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5710

## Description
Somehow adding horizontal margin broke layout. Replaced with padding.

## Screenshots/screencasts
https://streamable.com/qkyo3f

## Backward compatibility

### DIFF
--- a/css/layout-css/agenda-style.upload.css
+++ b/css/layout-css/agenda-style.upload.css
@@ -1600,10 +1600,18 @@
     padding: 6px 12px;
   }
 
+  .fl-with-top-menu .new-agenda-list-container {
+    width: 100%;
+  }
+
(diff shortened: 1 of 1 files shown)


tonytlwu:
@maksim-an-upplabs The beginning of the clip showed that the UI is clipped on the right when the agenda is vertically scrollable.

image

This doesn't look right.

When you switched to Preview, all the examples you've shown didn't show any scrollbar.

MaksymShokin:
@tonytlwu 
Now it looks correct. IE does not calculate vw properly.
ie-width

PR #366, opened 2020-04-16
IE agenda looks correct on all menu types

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5710#issuecomment-611412999

## Description
I have not realized that .fl-with-top-menu affects only default menu. This way it should work as expected. I tested with all menus.

## Screenshots/screencasts
ie-menus

## Backward compatibility

### DIFF
--- a/css/layout-css/agenda-style.upload.css
+++ b/css/layout-css/agenda-style.upload.css
@@ -1600,21 +1600,21 @@
     padding: 6px 12px;
   }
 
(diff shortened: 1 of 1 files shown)


tonytlwu:

Risk: Medium
Affected: LFD Agenda layout on IE 11
Verify: All UI controls and scrollbars are visible and correctly laid out when enabling different combinations of features.

PR #375, opened 2020-04-30
[Fix] Let Fliplet.Navigate.url() determine if authentication is needed

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-studio/issues/6318

Since this JS API change https://github.com/Fliplet/fliplet-api/pull/4032 the `Fliplet.Navigate.url()` self-determines whether the URL needs to be authenticated. There's therefore no need for the LFD to authenticate if it's passing the URL to `Fliplet.Navigate.url()`.

However, perhaps the fix should lie in the `Fliplet.Media.authenticate()` function so that the function never authenticates unless it's a native app? i.e. adding these 3 lines:

image

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -1278,7 +1278,6 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     }
 
     if (options.summaryLinkAction.type === 'url') {
-      value = Fliplet.Media.authenticate(value);
       Fliplet.Navigate.url(value);
     } else {
       var opt = { transition: 'fade' };


squallstar:
@tonytlwu I think `web` needs to keep using `Fliplet.Media.authenticate` because media files will most likely be under the CDN domain, which doesn't forward cookies and needs the auth token to be in the GET request.

tonytlwu:
@squallstar Perhaps `Fliplet.Media.authenticate()` should be updated so that it only authenticates if it's a native app **or** the URL is a CDN one. Currently, it appears that the function is authenticating API URLs as well.

squallstar:
@tonytlwu I still don't understand what's the problem with authenticating API URLs. Keep in mind that they may need authentication when accessing a webapp on apps.fliplet.com since you may not have any specific cookie on the APIs.

tonytlwu:
@squallstar This is the problem I see:

1. A user opens an `api.fliplet.com` URL in Studio preview
1. This is authenticated into a `cdn.api.fliplet.com` URL with the Studio user session token
1. The URL can now be shared with anyone to access the file, which is a security flaw
1. Opening the `api.fliplet.com` URL worked just fine without the `cdn.*` part and without any `auth_token`

If the web app is on `apps.fliplet.com` and won't have access to `api.fliplet.com`, perhaps the rule could be:

1. Don't authenticate if you're on a web preview; or
1. Don't authenticate if you're on `api.fliplet.com`?

PR #363, opened 2020-04-13
Limit text now is hidden when user is searching or filtering

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6186

## Description
Changed hidden class assignment.

## Screenshots/screencasts
https://streamable.com/39eya5

## Backward compatibility
This change is fully backward compatible.

## Reviewers 
@upplabs-alex-levchenko @YaroslavOvdii

PR #378, opened 2020-05-12
[Fix] Limit text now is hidden when user is searching or filtering

MaksymShokin:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6186

## Description
I remembered why I did not use "truncated" variable. We cannot use it because when we limit the number of searches truncated variable becomes true. Therefore is it always true when searches are limited.

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -1643,7 +1643,7 @@ DynamicList.prototype.searchData = function(options) {
 
       if (limitEntriesEnabled) {
         // Do not show limit text when user is searching or filtering
-        _this.$container.find('.limit-entries-text').toggleClass('hidden', !!results.filteredResult);
+        _this.$container.find('.limit-entries-text').toggleClass('hidden', results.filteredResult);
       }
 
       if (!_this.data.forceRenderList
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -1446,7 +1446,7 @@ DynamicList.prototype.renderLoopHTML = function () {
         // Changing close icon in the fa-times-thin class for windows 7 IE11
         if (/Windows NT 6.1/g.test(navigator.appVersion) && Modernizr.ie11) {
           $('.fa-times-thin').addClass('win7');
-        }        
+        }
 
         resolve(data);
       }
@@ -1643,7 +1643,9 @@ DynamicList.prototype.searchData = function(options) {
 
       if (limitEntriesEnabled) {
         // Do not show limit text when user is searching or filtering
-        _this.$container.find('.limit-entries-text').toggleClass('hidden', truncated);
+        var showLimitText = _this.isSearching || _this.isFiltering || _this.showBookmarks;
+        
+        _this.$container.find('.limit-entries-text').toggleClass('hidden', showLimitText);
       }
 
       if (!_this.data.forceRenderList
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -1643,9 +1643,7 @@ DynamicList.prototype.searchData = function(options) {
 
       if (limitEntriesEnabled) {
         // Do not show limit text when user is searching or filtering
-        var showLimitText = _this.isSearching || _this.isFiltering || _this.showBookmarks;
-        
-        _this.$container.find('.limit-entries-text').toggleClass('hidden', showLimitText);
+        _this.$container.find('.limit-entries-text').toggleClass('hidden', results.filteredResult);
       }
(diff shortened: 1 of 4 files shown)


tonytlwu:
@MaksymShokin @sofiiakvasnevska The prefix is either `[Fix]` or `[Feature]`, not `Bug.`

tonytlwu:
@MaksymShokin Your description suggests that `truncated` isn't a property that you can use. However, the fix isn't necessarily to add repeated logic that falls outside of the utility function that generates the `truncated` property.

Does it work if you update https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/a3bb49bcf81a1ba70aac0c759e32f497cc595f8f/js/utils.js#L537 to the below?

```js
truncated: truncated && !showBookmarks && _.isEmpty(activeFilters) && !values
```

MaksymShokin:
@tonytlwu 
We use a truncated variable in a few places across some files. I decided not to change it to not break logic in other places.
I tried your suggestion and it does not seem to work.

tonytlwu:
@MaksymShokin Then perhaps add a new `filteredResult` property to the response that equals `showBookmarks || !_.isEmpty(activeFilters) || values`. You can then use that property.

MaksymShokin:
@tonytlwu 
Fixed.

js/layout-javascript/news-feed-code.js
```diff
@@ -1643,7 +1643,7 @@ DynamicList.prototype.searchData = function(options) {
 
       if (limitEntriesEnabled) {
         // Do not show limit text when user is searching or filtering
-        _this.$container.find('.limit-entries-text').toggleClass('hidden', truncated);
+        _this.$container.find('.limit-entries-text').toggleClass('hidden', results.filteredResult);
```
tonytlwu, line 1646: Shouldn't `truncated` be used here in some way as well? Otherwise the message would be shown when there are less entries than the limit in the initial render.

```suggestion
        _this.$container.find('.limit-entries-text').toggleClass('hidden', !truncated || results.filteredResult);
```

MaksymShokin: @tonytlwu 
Actually it works fine without truncated variable. During testing found another bug, made commit to fix it. Should work correctly now.

js/utils.js
```diff
@@ -544,7 +544,8 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
 
     return Promise.resolve({
       records: searchResults,
-      truncated: truncated
+      truncated: truncated,
+      filteredResult: showBookmarks || !_.isEmpty(activeFilters) || value
```
tonytlwu, line 548: Cast this into a boolean value instead of having to cast it elsewhere. 

```suggestion
      filteredResult: !!(showBookmarks || !_.isEmpty(activeFilters) || value)
```

js/layout-javascript/news-feed-code.js
tonytlwu: @MaksymShokin Can you explain why it isn't needed?

js/utils.js
MaksymShokin: Fixed

js/layout-javascript/news-feed-code.js
MaksymShokin: @tonytlwu 
I tested it and it works without truncated.

tonytlwu: @MaksymShokin Yes, that sounds like what you said previously. I needed an explanation for why it isn't needed, not the results. It needs to be explained because before your changes in https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/363/files `truncated` was part of the logic. Because you're now effectively removing it from the logic, it needs to be explained when asked.

Here's an example.

I took a look at your changes, tested it myself and looked into the code. I found that [this code](https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/a3bb49bcf81a1ba70aac0c759e32f497cc595f8f/js/layout-javascript/news-feed-code.js#L1552-L1555) suggests that `limit` is only a valid number (above `-1`) if search, filter and bookmarks are disabled.

This suggests that `results.truncated` will never be `true` if search/filter/bookmarks are enabled. It also means that the solution from the beginning could just be:

```js
_this.$container.find('.limit-entries-text')[truncated && limit > 0 ? 'removeClass' : 'addClass']('hidden');
```

This could work because the variable `limit` has already taken the search, filter and bookmark states into consideration. It's a simpler solution, requires no new logic wither in the layout JS or the `utils.js` files.

What do you think? Would this work, and do you have any reasons for preferring other solutions?

PR #382, opened 2020-05-14
[Feature] Changed opacity to rgba so that component could change normally through new appearance settings. 

YaroslavOvdii:
@sofiiakvasnevska 

## Description
In this PR we are changing only `opacity` to `rgba` there is no visual changes happens.

## Backward compatibility

## Reviewers 
@upplabs-alex-levchenko

## Notes

These changes are needed for this [feature](https://github.com/Fliplet/fliplet-studio/issues/6225) to work right.

### DIFF
--- a/css/layout-css/small-card-style.upload.css
+++ b/css/layout-css/small-card-style.upload.css
@@ -726,8 +726,7 @@
 }
 
 .new-small-card-list-container .small-card-list-location {
-  color: #333333;
-  opacity: 0.5;
+  color: rgba(51, 51, 51, 0.5);
   font-size: 12px;
   font-weight: 500;
   line-height: 1.6;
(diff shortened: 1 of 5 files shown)


css/layout-css/small-card-style.upload.css
```diff
@@ -985,7 +982,6 @@
 .small-card-detail-overlay .small-card-list-detail-button a:active,
 .small-card-detail-overlay .small-card-list-detail-button a:active:hover,
 .small-card-detail-overlay .small-card-list-detail-button a:active:focus {
-  color: #333333;
```
hcarneiro, line 988: Are we sure that removing this will not make the colour of the link some other colour when hovered, in focus, and/or active?

YaroslavOvdii: @hcarneiro It doesn't change color on hover or active on link.
https://share.getcloudapp.com/7Ku0nzkm

PR #389, opened 2020-06-09
[Fix] Show correct error message when the user doesn't have permissions to the Data Source.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6548

## Description
Show correct error message when the user doesn't have permissions to the Data Source.

## Screenshots/screencasts
image

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

## Notes
Removed `.catch()` from `connectToDataSource()` function give controll over error handlers to the `initialize()` function.
It wouldn't have any effect on the current flow of the Agenda layout because the `connectToDataSource()` function used only with the `initialize()` function.

### DIFF
--- a/js/layout-javascript/agenda-code.js
+++ b/js/layout-javascript/agenda-code.js
@@ -1040,10 +1040,6 @@ DynamicList.prototype.connectToDataSource = function() {
     }
 
     return getData(cache);
-  }).catch(function (error) {
-    Fliplet.UI.Toast.error(error, {
-      message: 'Error loading data'
-    });
   });
 }
 


YaroslavOvdii:
@tonytlwu not quite if you look at this [video](https://share.getcloudapp.com/P8ueWdBE) you may see that we are sowing error toast twice. 
The first time it comes from the `connectToDataSource` function and we may see the error from the back end at that error message but it overrides by next error coming from this method `renderDatesHTML` at this [line](https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/master/js/layout-javascript/agenda-code.js#L1281)

PR #390, opened 2020-06-09
[Feature] Added filter options for checking if the value is empty or not

MaksymShokin:
@sofiiakvasnevska

## Issue
https://github.com/Fliplet/fliplet-studio/issues/4800

## Description
Added possibility to filter by empty or not empty value. Value field will be hidden if empty or not empty field is selected.

## Screencast
https://streamable.com/usc02a

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/utils.js
```diff
@@ -394,6 +394,14 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
 
     return _.filter(records, function (record) {
       return _.every(filters, function (filter) {
+        if (filter.condition === 'empty') {
+          return _.isEmpty(record.data[filter.column]);
```
tonytlwu, line 398: Numbers will be incorrectly reported as empty using `_.isEmpty()`.

Also, `NaN` is reported as a number using `_.isNumber()`.

This means:

```suggestion
          return _.isEmpty(record.data[filter.column]) && (_.isNaN(record.data[filter.column]) || !_.isNumber(record.data[filter.column]));
```

Also, use variables laid out in https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/feat/filter-empty/js/utils.js#L397-L400 which will help simplify the code.

MaksymShokin: Fixed. I used isFinite lodash method

js/utils.js
```diff
@@ -394,28 +394,33 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
       return _.every(filters, function (filter) {
-        if (filter.condition === 'empty') {
-          return _.isEmpty(record.data[filter.column]);
-        }
-          
-        if (filter.condition === 'notempty') {
-          return !_.isEmpty(record.data[filter.column]); 
-        }
+        var condition = filter.condition;
+        var rowData = _.get(record, 'data.' + filter.column, null);
```
tonytlwu, line 398: Using a string path ends up using `String.prototype.split()` which does not perform very well. This isn't usually a problem, but on an operation like this where the operation is executed over a potentially very large list, it's more efficient to pass an array as path.

```suggestion
        var rowData = _.get(record, ['data', filter.column], null);
```

js/utils.js
tonytlwu: Didn't know about `_.isFinite()`. Thanks for teaching me something!

js/utils.js
MaksymShokin: Fixed

MaksymShokin:
@tonytlwu 
I tested all optioned you have mentioned. Now it should work correctly.

js/utils.js
```diff
@@ -394,20 +394,33 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+
+        if (condition === 'none' || filter.column === 'none') {
           // Filter isn't configured correctly
           return true;
         }
 
-        var condition = filter.condition;
-        var rowData;
+        if (condition === 'empty') {
+          return _.isEmpty(rowData) && !_.isFinite(rowData) && rowData !== true && rowData !== false;
```
tonytlwu, line 406:
```suggestion
          return _.isEmpty(rowData) && !_.isFinite(rowData) && typeof rowData !== 'boolean';
```

MaksymShokin: Fixed

PR #391, opened 2020-06-09
[Feature] Show data sources connected to the app first in the list.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/3164

## Description
Show data sources connected to the app first in the list.

## Screenshots/screencasts
https://share.getcloudapp.com/mXuAjBjm

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -613,7 +613,7 @@ var DynamicLists = (function() {
                   _this.config.dataAlertSeen = true;
                 }
 
-                allDataSources = dataSources;
+                allDataSources = _this.sortDataSourcesListByCurrentAppId(dataSources);
                 _this.getColumns(_this.config.dataSourceId);
                 _this.initSelect2(allDataSources);
                 _this.initSecondSelect2(allDataSources);
@@ -621,7 +621,7 @@ var DynamicLists = (function() {
               });
             }
 
-            allDataSources = dataSources;
+            allDataSources = _this.sortDataSourcesListByCurrentAppId(dataSources);
             _this.getColumns(_this.config.dataSourceId);
             _this.initSelect2(allDataSources);
             _this.initSecondSelect2(allDataSources);
@@ -1193,6 +1193,27 @@ var DynamicLists = (function() {
       _this.handleTokensSelection();
       _this.loadTokenFields();
     },
+    sortDataSourcesListByCurrentAppId: function(dataSources) {
+      var currentAppDataSourcesId = dataSources.reduce(function(connectedDSIds, dataSource) {
+        if (dataSource.appId === appId) {
+          connectedDSIds.push(dataSource.id);
+        }
+
+        dataSource.apps.find(function(app) {
+          if (app.id === appId) {
+            connectedDSIds.push(dataSource.id);
+            return true;
+          }
+        });
+        
+        return connectedDSIds;
+      }, []);
+
+      return dataSources.sort(function(a, b) {
+        // Check if DS connected to the current app
+        return currentAppDataSourcesId.indexOf(a.id) !== -1 && currentAppDataSourcesId.indexOf(b.id) === -1 ? -1 : 1;
+      });
+    },
     removeFocusFromTokenInput: function () {
       $('input.token-input.ui-autocomplete-input').blur();
     },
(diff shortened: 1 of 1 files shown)


YaroslavOvdii:
@tonytlwu Yes, it is possible.
In order, these changes worked properly I had to change `customDsSearch` because we have changed the structure of the select2 data.

js/interface.js
```diff
@@ -1194,25 +1194,36 @@ var DynamicLists = (function() {
       _this.loadTokenFields();
     },
     sortDataSourcesListByCurrentAppId: function(dataSources) {
-      var currentAppDataSourcesId = dataSources.reduce(function(connectedDSIds, dataSource) {
+      var groupedDataSources = [
+        {
+          id: 'currentAppDSs',
+          text: 'This APP',
```
tonytlwu, line 1200:
```suggestion
          text: 'This app',
```

js/interface.js
```diff
@@ -1194,25 +1194,36 @@ var DynamicLists = (function() {
...
+          name: 'currentApp',
+          children: []
+        },
+        {
+          id: 'OtherDSs',
+          text: 'Other APPs',
```
tonytlwu, line 1206:
```suggestion
          text: 'Other apps',
```

YaroslavOvdii:
@tonytlwu sorry though I've added it.
[Screencast](https://share.getcloudapp.com/6qu22yxA)
Made required changes.

js/interface.js
```diff
@@ -1193,6 +1193,38 @@ var DynamicLists = (function() {
...
+        {
+          id: 'OtherDSs',
+          text: 'Other apps',
+          name: 'otherApp',
+          children: []
+        }
+      ];
+
+      dataSources.forEach(function(dataSource) {
+        if (dataSource.appId === appId) {
```
tonytlwu, line 1213: @YaroslavOvdii This suggests that we're using `dataSource.appId` to determine if the data source is associated with an app.

In the following scenario:

1. A data source is created, but not for the app
1. The data source is associated with the app using the developer option feature in Studio

In this case, is the data source also added to the `This app` list?

js/interface.js
```diff
@@ -1193,6 +1193,38 @@ var DynamicLists = (function() {
       _this.handleTokensSelection();
       _this.loadTokenFields();
     },
+    sortDataSourcesListByCurrentAppId: function(dataSources) {
+      var groupedDataSources = [
+        {
+          id: 'currentAppDSs',
+          text: 'This app',
+          name: 'currentApp',
+          children: []
```
tonytlwu, line 1202: In what order are the data sources listed within each group?

js/interface.js
```diff
@@ -1193,6 +1193,38 @@ var DynamicLists = (function() {
       _this.handleTokensSelection();
       _this.loadTokenFields();
     },
+    sortDataSourcesListByCurrentAppId: function(dataSources) {
+      var groupedDataSources = [
+        {
+          id: 'currentAppDSs',
+          text: 'This app',
+          name: 'currentApp',
```
tonytlwu, line 1201: `id` and `name` seem to serve very similar purpose here. Can we just keep one?

I prefer the value `currentApp` more than `currentAppDSs`.

js/interface.js
YaroslavOvdii: @tonytlwu 

> A data source is created, but not for the app

If the data source doesn't have an `appId` or `apps` array doesn't contain current appId.
This data source will be added to the `Other apps` list.

> The data source is associated with the app using the developer option feature in Studio

And if a situation will be opposite to described above we will add this data source to the `This app` list.

js/interface.js
YaroslavOvdii: @tonytlwu it is not sorted at all. I will add sorting as you have described.

js/interface.js
YaroslavOvdii: > id and name seem to serve very similar purpose here. Can we just keep one?

We can, but we will have to alter a little bit more the `customDSsearch` function so it worked correctly. Because now we show entries only if they have `id`, `text` and `name` keys in it.
Should we do that?

> I prefer the value currentApp more than currentAppDSs.
I will change it.

PR #399, opened 2020-07-06
[Feature] Added possibility to upload multiple images

MaksymShokin:
@sofiiakvasnevska

## Issue
https://github.com/Fliplet/fliplet-studio/issues/4865

## Description
I made PR only for the directory, if PR will be approved then I will make other layouts.

## Screencast
https://streamable.com/sgegno

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


templates/build/small-card-detail.build.hbs
```diff
@@ -52,13 +52,22 @@
...
+                <div class="small-card-list-detail-label">\{{label}}</div>
+              \{{/if}}
+              <div class="small-card-list-detail-body-text">
+                <img src="\{{validateImage this}}">
+              </div>
+            \{{/each}}
+          \{{else}}
+            <div class="small-card-multiple-images-holder">
+              \{{#each content}}
+                <div class="small-card-list-multiple-images-item" id="small-card-photoswipe">
```
tonytlwu, line 67: This suggests there will be multiple DOM elements on the same with the same ID. We can't have that.

js/layout-javascript/small-card-code.js
```diff
@@ -1623,6 +1646,14 @@ DynamicList.prototype.addDetailViewData = function (entry) {
       content = entry.originalData[dynamicDataObj.column];
     }
 
+    if (dynamicDataObj.type == 'image') {
+      content = entry.originalData[dynamicDataObj.column].split(/\n/);
+
+      content.forEach(function(imageUrl) {
```
tonytlwu, line 1652: Can we not use `.map` here?

MaksymShokin: @tonytlwu 
We actually need "content" variable. We use it later. Therefore we can use map instead of forEach, but I do not see any benefit of using it.

MaksymShokin:
@tonytlwu 
I added other layouts. So new approve is required

js/layout-javascript/small-h-card-code.js
```diff
@@ -335,6 +342,22 @@ DynamicList.prototype.attachObservers = function() {
       }).then(function() {
         Fliplet.UI.Actions(options);
       });
+    })
+    .on('click', '.small-h-card-multiple-images-item', function() {
+      _this.imagesData.options.index = $(this).index();
+
+      var smallHCardImageGallery = Fliplet.Navigate.previewImages( _this.imagesData);
```
tonytlwu, line 349:
```suggestion
      var smallHCardImageGallery = Fliplet.Navigate.previewImages(_this.imagesData);
```

js/layout-javascript/small-card-code.js
```diff
@@ -61,6 +65,9 @@ function DynamicList(id, data, container) {
 
   this.data.bookmarksEnabled = _this.data.social.bookmark;
 
+  // this is used for calling photoswipe library, do not remove
+  var photoswipeTemplate = Fliplet.Widget.Templates['templates/photoswipe.photoswipe'];
```
tonytlwu, line 69: In what way is this used?

`Fliplet.Navigate.previewImages()` has a similar line but it's being used within the function.

image

It doesn't look like `photoswipeTemplate` is being used at all.

MaksymShokin: Actually you are right. Although before it did not work without it, maybe there was another mistake. I removed photoswipe vars.

js/layout-javascript/small-h-card-code.js
MaksymShokin: Fixed

PR #406, opened 2020-07-16
[Feature] Allow users to use the keyboard to get access to the main features of the LFD.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6423

## Description
Allow users to use the keyboard to get access to the main features of the LFD.

## Screenshots/screencasts
https://share.getcloudapp.com/z8u82qKx

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/layout-javascript/simple-list-code.js
```diff
@@ -276,6 +292,7 @@ DynamicList.prototype.attachObservers = function() {
       $parentElement.find('.hidden-filter-controls').addClass('active');
       $parentElement.find('.list-search-cancel').addClass('active');
       $elementClicked.addClass('active');
+      $parentElement.find('[data-filter-group]').prop('hidden', null);
```
tonytlwu, line 295: I usually use `.prop(key, false)` to disable a property, not `.prop(key, null)`. Can you confirm if there's a difference?

YaroslavOvdii: In this case, there is no difference I simply took an example from this [article](https://api.jquery.com/removeAttr/)

PR #396, opened 2020-06-25
[Feature] Showing an active filters that users selected.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5641

## Description
Showing an active filters that users selected.

## Screenshots/screencasts
https://share.getcloudapp.com/04uYx5Gb

## Backward compatibility
This change is fully backward compatible.

## Reviewers
@upplabs-alex-levchenko

## Notes
When user clicks on an active filter it will be deselected.

PR #423, opened 2020-08-25
[Feature] Showing an active filters that users selected.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5641

## Description
Showing an active filters that users selected.

## Screenshots/screencasts
https://share.getcloudapp.com/04uYx5Gb

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

## Notes
When user clicks on an active filter it will be deselected.


tonytlwu (changes requested):
@YaroslavOvdii I know this is the work based on https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/396, which is based on the snippet shared in https://github.com/Fliplet/fliplet-studio/issues/5641#issuecomment-627882622

On second thought, the snippet was designed that way because developers can only use the available hooks to mock a certain behaviour.

If active filters are to be shown to users, we should use the `flListDataAfterRenderList` hook for two reasons:

1. We'd like to show the active filters before the list starts rendering
1. I'd eliminate using hooks because if the hook execution was paused/executed, we still want as much of our core features to work as possible

YaroslavOvdii:
@tonytlwu 

> If active filters are to be shown to users, we should use the flListDataAfterRenderList hook for two reasons:

But we already using this hook, maybe you mean that we need flListDataBeforeRenderList?

tonytlwu:
Sorry, I meant we should NOT use the hook...

YaroslavOvdii:
Resovled confilcts in the new [PR](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/429)

PR #425, opened 2020-08-25
[Feature] Added possibility to upload multiple images

MaksymShokin:
@sofiiakvasnevska

## Issue
Fliplet/fliplet-studio#4865

## Description
Remade PR after reverting and fixed an issue mentioned by Inna.

## Screencast
https://streamable.com/sgegno

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


tonytlwu:
@MaksymShokin See https://github.com/Fliplet/fliplet-studio/issues/4865#issuecomment-680789393

MaksymShokin:
@tonytlwu 
Added new styling

MaksymShokin:
mobile
desktop

css/layout-css/agenda-style.upload.css
```diff
@@ -1525,18 +1525,43 @@
...
+  border: 1px solid rgba(127, 127, 127, 0.5);
   cursor: pointer;
 }
 
+.agenda-multiple-images-item:nth-of-type(2n+1) {
+  margin-right: 15px;
+}
+
+.agenda-multiple-images-item img {
+  object-fit: contain;
```
tonytlwu, line 1549: Does this work in IE 11? https://caniuse.com/#feat=object-fit says it doesn't.

css/layout-css/agenda-style.upload.css
```diff
@@ -1525,18 +1525,43 @@
...
 .agenda-multiple-images-item {
-  flex: 1 1 70px;
-  margin: 10px;
-  border-radius: 5px;
-  overflow: hidden;
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  justify-content: center;
+  width: 149px;
```
tonytlwu, line 1535: How is this making sure that it's always presented as a 2 or 3 column layout in squares if the width and height are set in absolute values?

MaksymShokin: How can we set it in percentage?

MaksymShokin: @tonytlwu 
To make sure that the column has 2 or 3 items per row we can add max-width to the images container in px. So, far this is the only way we have found. Is it ok ?

tonytlwu: @MaksymShokin I don't understand what the challenge is here. There are many methods in CSS to create a 2 or 3 column layout.

MaksymShokin: @tonytlwu 
Columns-count property does not work properly here due to wrong ordering. Could you share other methods you have in mind? I can use css grid. But we still need to set absolute values for height and width.

tonytlwu: https://jsfiddle.net/tonytlwu/7r26b8cz/

MaksymShokin: I did not get how is jsfiddle above is going to help. We have 1 array of elements that we have to sort in 2 or 3 items per row.

tonytlwu: @MaksymShokin My mistake. I hadn't tested with more than 3 items. See https://jsfiddle.net/tonytlwu/7r26b8cz/ again now.

MaksymShokin: @tonytlwu 
I do not see any changes. If I add fourth item it does not go onto the next row.

tonytlwu: @MaksymShokin Sorry. The link was incorrect. See https://jsfiddle.net/tonytlwu/7r26b8cz/3

MaksymShokin: @tonytlwu 
In the requirements, it was said that containers should be squares. You added width: 33.33%. How can we add the same value to the height?

tonytlwu: Use either one of these principles https://jsfiddle.net/tonytlwu/bL2hkcae/

MaksymShokin: Fixed

js/layout-javascript/agenda-code.js
```diff
@@ -744,6 +748,22 @@ DynamicList.prototype.attachObservers = function() {
...
       record.bookmarkButton.like();
+    })
+    .on('click', '.agenda-multiple-images-item', function() {
+      _this.imagesData.options.index = $(this).index();
+
+      var agendaImageGallery = Fliplet.Navigate.previewImages(_this.imagesData);
+
+      agendaImageGallery.listen('afterChange', function() {
+        Fliplet.Page.Context.update({
+          agendaImageGalleryId: _this.data.id,
```
tonytlwu, line 759: Why are we updating these page context parameters? Does this feature support opening a detail view gallery item using URL parameters? If so, I don't see any code that supports this.

tonytlwu (changes requested):

widget.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "List (from data source)",
-  "package": "com.fliplet.dynamic-lists",
+  "name": "List (from data source)-images",
+  "package": "com.fliplet.dynamic-lists-images",
```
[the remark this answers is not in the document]
MaksymShokin: fixed

js/layout-javascript/agenda-code.js
MaksymShokin: That is how it is done in the gallery component. I do not think that is it possible to open a detail view using URL, as far as I remember this is done to correctly assign images to specific component. So if we have multiple gallery components it will not show all pictures from both of those components.
images

tonytlwu: If the page context updating isn't going to do anything for the user, then take it out. It wasn't part of the requirement to get that working.

tonytlwu (changes requested):

MaksymShokin:
@tonytlwu 
If I remove it then fullscreen mode does not work for images.

tonytlwu:
@MaksymShokin Can you explain or show what fullscreen mode you're referring to, and why it doesn't work without the page context?

MaksymShokin:
@tonytlwu 
By fullscreen mode I mean this.
fullscreen

MaksymShokin:
@tonytlwu 
I do not know why it does not work without it, I copied it from the gallery. Actually you wrote this code in gallery, so I think there is should be logic behind it.
gallery

tonytlwu:
@MaksymShokin **You need to find out why it doesn't work without the page context.** As far as I know they shouldn't be necessary to the working of the feature.

By setting the page context, the URL query parameter, e.g. `?agendaImageGalleryId=3`, gets updated. These are present only to make sure that when a page loads, it knows to process the query and take users to a specific state of the page.

Without it, gallery feature should work just fine.

PR #428, opened 2020-08-31
[Fix] Show add button at 20 px right from the LFD item list.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-18?atlOrigin=eyJpIjoiYmY0ZDUzNzFlZTY4NDU0ZDgwNjdiMTYyMzc3NmEzMjkiLCJwIjoiaiJ9

## Description
Show the add button at 20 px right from the LFD item list.

## Screenshots/screencasts
https://share.getcloudapp.com/mXuybWxO

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -394,6 +394,36 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     }
   }
 
+  function moveAddbuttonPosition(instance) {
+    var $addButton = instance.$container.find('.dynamic-list-add-item');
+    var layout = instance.data.layout;
+    var listClasses = {
+      'agenda': '.agenda-list-card-holder',
+      'news-feed': '.news-feed-list-wrapper',
+      'simple-list': '.simple-list-wrapper',
+      'small-card': '.small-card-list-wrapper'
+    };
+    var elementSpace = 20;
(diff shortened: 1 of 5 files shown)


js/utils.js
```diff
@@ -394,6 +394,36 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     }
   }
 
+  function moveAddbuttonPosition(instance) {
```
tonytlwu, line 397: Always use an options object with the utility functions because users can modify and customise layout JS and it would be a lot harder to extend the util functions later on if the customised layout JS don't assume an options parameter.

```suggestion
  function moveAddbuttonPosition(options) {
```

tonytlwu (changes requested):

js/utils.js
YaroslavOvdii: @tonytlwu Fixed

tonytlwu (approved):

Risk: Medium
Affected: Add buttons in LFD layouts
Verify: Add buttons are correctly positioned on all device sizes

PR #442, opened 2020-10-06
OD-5 [Feature] LFD Accessibility

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-5 https://weboo.atlassian.net/jira/software/projects/OD/boards/9?selectedIssue=OD-5
## Description
Added accessility for all LFD components
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cY6h1oKrmL
https://screencast-o-matic.com/watch/cY6h1bKrHi
https://screencast-o-matic.com/watch/cY6h1YKrHq
https://screencast-o-matic.com/watch/cY6h1rKrHw
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


templates/build/small-card-base.build.hbs
```diff
@@ -52,14 +58,14 @@
...
+      <div class="toggle-bookmarks focus-outline" tabindex="0" data-mixitup-control="\{{id}}" data-toggle=".bookmarked" >
         <span class="fa fa-bookmark-o"></span>
         <span class="fa fa-bookmark"></span>
       </div>
       \{{/if}}
-      <div class="list-search-cancel" ><span class="fa fa-times"></span></div>
+      <div class="list-search-cancel focus-outline" tabindex="0" ><span class="fa fa-times"></span></div>
       \{{#if filtersEnabled}}
-      <i class="fa fa-sliders" ></i>
+      <i class="fa fa-sliders focus-outline" tabindex="0" ></i>
```
tonytlwu, line 68: No spaces before `>`. Check all templates.

```suggestion
      <i class="fa fa-sliders focus-outline" tabindex="0"></i>
```

tonytlwu (changes requested):

js/utils.js
```diff
@@ -363,6 +363,16 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     }
   }
 
+  function accessibilityDetails(event, item) {
```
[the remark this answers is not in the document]
sofiiakvasnevska: this function has been reverted in one of the previous PRs

js/layout-javascript/simple-list-code.js
```diff
@@ -73,17 +73,17 @@ function DynamicList(id, data, container) {
...
     if (_.get(session, 'entries.dataSource.data')) {
-       _.extend(_this.myUserData, _.get(session, 'entries.dataSource.data'));
+      _.extend(_this.myUserData, _.get(session, 'entries.dataSource.data'));
     }
 
     // Start running the Public functions
     _this.initialize();
   });
-};
+}
```
tonytlwu, line 82: Find all other instances. I don't want to look through the PR to try and pick them all out.

```suggestion
};
```

tonytlwu (changes requested):

js/utils.js
```diff
@@ -363,6 +363,16 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+    if (event.key === 'Enter' || event.type === 'click') {
+      if (event.type === 'click') {
+        item.removeClass('focus-outline');
```
tonytlwu, line 369: Why does the `focus-outline` class need to be removed?

js/utils.js
```diff
@@ -363,6 +363,16 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
...
+      }
+      return true;
```
tonytlwu, line 371: Empty lines before and after `{ }` except when it's in the beginning or end of a `{ }` block.

```suggestion

      return true;
```

tonytlwu (changes requested):

js/utils.js
AndrRyaz: because when the user clicks on the element with the mouse, this class is also activated as I understand it, we only need to add the class when the keyboard event is pressed

js/utils.js
AndrRyaz: Done

js/layout-javascript/simple-list-code.js
AndrRyaz: I went through all the files and double-checked the presence of ';'

templates/build/agenda-base.build.hbs
```diff
@@ -30,6 +30,12 @@
   \{{/if}}
   <!-- End of Filter Overlay -->
 
+  \{{#if showAddEntry}}
+  <div class="dynamic-list-add-item focus-outline" tabindex="0">
```
tonytlwu, line 34: Why did the location of this `.dynamic-list-add-item` DOM element need to be moved? We have existing CSS selectors in this component and in the theme SCSS file that relies on the current structure.

js/utils.js
tonytlwu: I believe we rely on the `focus-outline` class to be available so that when user focusses on the UI using **TAB**, the UI will be correctly outlined according to the theme settings.

This means that if `focus-outline` is removed after a user clicks on the UI, then subsequent focuses using **TAB** will not be correctly outlined.

tonytlwu (changes requested):

templates/build/agenda-base.build.hbs
AndrRyaz: The structure of the page shows that the button for creating a new record is in the foreground and its selection should be earlier than the elements below it.  Or I'm wrong?

js/utils.js
AndrRyaz: Yes, your right. I can remove this code but then we need to make changes in the logic of the `focus-outline` class, because this class has pseudo class focus-within and it shows border when event click

templates/build/agenda-base.build.hbs
tonytlwu: I see, so you're moving this to make sure the order of selection puts the **Add** button before the list items. I think that's a good consideration but we don't need to do that for now.

AndrRyaz: Ok, i can revert this

js/utils.js
AndrRyaz: @tonytlwu Should I remove this part of the code?

tonytlwu: @AndrRyaz I'm reviewing the current solution and class names used to manage focus outlines. I'll let you know the results when I'm done cc @sofiiakvasnevska

templates/build/agenda-base.build.hbs
tonytlwu: Yes please. And there's also conflict with this PR so you may want to resolve them first.

js/layout-javascript/small-h-card-code.js
```diff
@@ -99,248 +101,258 @@ DynamicList.prototype.attachObservers = function() {
...
-                container: _this.$container
-              })
-                .then(function() {
-                  if (_this.data.deleteData && typeof _this.data.deleteData === 'function') {
-                    return _this.data.deleteData(entryID);
-                  }
-
-                  return _this.deleteEntry(entryID);
+    .on('click keydown', '.dynamic-list-delete-item', function(event) {
+      if (_this.Utils.Accessibility.accessibilityDetails(event, $(this))) {
```
tonytlwu, line 279: 1. You're sending a variable that isn't used.
2. Change the function name and namespace.
3. Use a `if (flag) { return } /* Continue  */` logic to reduce the need for indentation

```suggestion
      if (!_this.Utils.Event.isExecEvent(event)) {
        return;
      }
```

js/utils.js
```diff
@@ -367,6 +367,14 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     }
   }
 
+  function accessibilityDetails(event) {
+    if (event.key === 'Enter' || event.type === 'click') {
```
tonytlwu, line 371:
```suggestion
    return event.key === 'Enter' || event.type === 'click';
```

tonytlwu (changes requested):

js/layout-javascript/small-h-card-code.js
tonytlwu: Number 2 was not actioned

tonytlwu: Number 2 was not actioned correctly. The namespace is incorrect.

js/layout-javascript/news-feed-code.js
```diff
@@ -174,8 +174,10 @@ DynamicList.prototype.attachObservers = function() {
       _this.hideFilterOverlay();
       _this.clearFilters();
     })
-    .on('click', '.hidden-filter-controls-filter', function() {
-
+    .on('click keydown', '.hidden-filter-controls-filter', function(event) {
+      if (!_this.Utils.accessibilityHelpers.isExecute(event)) {
+        return;
+      }
```
tonytlwu, line 180: An empty line of spacing before and after `{ }` unless its in the beginning of a `{ }` block.

@upplabs-alex-levchenko @sofiiakvasnevska Please make sure Upplabs reviews all the PRs and review notes correctly. This PR has taken too long to get right just because the usual coding standards are not being followed. We should not be picking it out another error like this in the next review.

```suggestion
      }

```

js/layout-javascript/news-feed-code.js
```diff
@@ -289,9 +308,13 @@ DynamicList.prototype.attachObservers = function() {
...
       _this.closeDetails();
     })
-    .on('click', '.list-search-icon .fa-sliders', function() {
+    .on('click keydown', '.list-search-icon .fa-sliders', function(event) {
+      if (!_this.Utils.accessibilityHelpers.isExecute(event)) {
+        return;
+      }
       var $elementClicked = $(this);
       var $parentElement = $elementClicked.parents('.new-news-feed-list-container');
+      _this.$container.find('[data-filter-group]').prop('hidden', null);
```
tonytlwu, line 317: An empty line of spacing before and after each `var` block unless its in the beginning of a `{ }` block.

cc @upplabs-alex-levchenko @sofiiakvasnevska

```suggestion

      _this.$container.find('[data-filter-group]').prop('hidden', null);
```

tonytlwu (changes requested):

PR #444, opened 2020-10-12
OD-9 [Feature] Integration DS Provider into LFD

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-9 https://weboo.atlassian.net/jira/software/projects/OD/boards/9?selectedIssue=OD-9
## Description
Integration DS Provider into LFD
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cY6lob7cuR
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


js/interface.js
```diff
@@ -94,6 +95,43 @@ var DynamicLists = (function() {
     _this.init();
   }
 
+  function initDataSourceProvider(currentDataSourceId) {
+    var dataSourceData = {
+      dataSourceTitle: 'Profile data source',
+      dataSourceId: currentDataSourceId,
+      appId: Fliplet.Env.get('appId'),
+      default: {
+        name: 'LFD 0000',
```
tonytlwu, line 104: This doesn't look like an appropriate name for the data source.

js/interface.js
```diff
@@ -94,6 +95,43 @@ var DynamicLists = (function() {
     _this.init();
   }
 
+  function initDataSourceProvider(currentDataSourceId) {
+    var dataSourceData = {
+      dataSourceTitle: 'Profile data source',
```
tonytlwu, line 100: The LFD doesn't need to select a **Profile data source**. This seems inappropriately labelled.

tonytlwu (changes requested):
image

I see in the video (above) that it's still using the wrong data source selector dropdown. I'd like this verified with the latest data source provider UI to confirm the integration is working correctly.

AndrRyaz:
@tonytlwu yes the video has an old version here is a new one 
https://screencast-o-matic.com/watch/cYXbliLic4

js/interface.js
```diff
@@ -94,6 +95,43 @@ var DynamicLists = (function() {
     _this.init();
   }
 
+  function initDataSourceProvider(currentDataSourceId) {
+    var dataSourceData = {
+      dataSourceTitle: 'Your list data',
+      dataSourceId: currentDataSourceId,
+      appId: Fliplet.Env.get('appId'),
+      default: {
+        name: 'List from data source - ' + Fliplet.Env.get('appName'),
```
tonytlwu, line 104: https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/4743c434ff975cece766f99c77487a02fb3725e6/js/interface.js#L1627

The current implementation gives you a default data source name based on a combination of the app name and the LFD layout.

We should keep the same functionality.

tonytlwu (changes requested):

js/interface.js
AndrRyaz: @tonytlwu  resolved, in the future I will use this structure

tonytlwu: @AndrRyaz Essentially: pay attention to how features work before you make the changes, and don't break them.

tonytlwu (changes requested):
0:34 in https://screencast-o-matic.com/watch/cYXbliLic4 shows that the column dropdown is empty after changing the data source, which appears like a bug. If not, there isn't a clear explanation for why it's OK for the dropdown to be empty.

AndrRyaz:
@tonytlwu here is a new video https://screencast-o-matic.com/watch/cYXY6ALTKi, 
In the previous video, the selected date source did not have any fields in it - therefore it did not show anything, I already added default fields when creating the date source

tonytlwu (approved):

Risk: High
Verify: All data source related features can be correctly configured to work as expected

Squash and merge: Integrate new data source pvovider

PR #461, opened 2020-12-11
OD-253 [Feature] Filter value type: User profile data

AndrRyaz:
@sofiiakvasnevska 
## Issue
https://weboo.atlassian.net/jira/software/projects/OD/boards/9?selectedIssue=OD-253
## Description
1. Studio users can configure LFD to filter entries based on values loaded from user profile data

2. Studio users cannot save the form if the key (Value for) is not provided when the user saves the filter with the value type User profile data

3. I can use this filter value type on any of the filter logic except for: is empty, is not empty and any of the date comparison logics
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYlXDj8fY3
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


tonytlwu (changes requested):
I don't see anything in the screencast or code that demonstrate the filters are using values from the app user's profile/session data.

AndrRyaz:
@tonytlwu yes, sorry, i already added the last commits and make some fixes with eslint

PR #462, opened 2020-12-11
OD-222 [Fix] Correct LFD entries sorting by number

ivandevupp:
@sofiiakvasnevska
## Issue
OD-222 https://weboo.atlassian.net/browse/OD-222
## Description
Fixed incorrect sort type recognition and adjusted for sorting by floats
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYlXD88fgw
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


js/utils.js
```diff
@@ -1049,8 +1049,8 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     var isSortAsc = options.sortOrder === 'asc';
     var sortField = options.sortField;
     var startsWithAlphabet = /^[A-Z,a-z]/;
-    var sortType = getFieldType(records[0][sortField]);
-  
+    var sortType = getFieldType(records[0].data ? records[0].data[sortField]: records[0][sortField]);
```
tonytlwu, line 1052: Why would the fields be nested inside a `data` object and how is that related to the issue OD-222?

js/utils.js
```diff
@@ -1081,11 +1081,11 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
 
           return 0;
         case 'number':
-          if (!parseInt(aValue, 10)) {
+          if (+aValue !== 0 && !parseFloat(aValue, 10)) {
```
tonytlwu, line 1084: Please explain your fix here. And, why do we need to use `+aValue` when we have `parseFloat()` already?

js/utils.js
```diff
@@ -1184,6 +1184,10 @@ Fliplet.Registry.set('dynamicListUtils', (function () {
     var timeRegex = /^[0-9]{1,2}:[0-9]{1,2}$/
     
     if (valueType === 'string') {
+      if (!isNaN(value)) {
```
tonytlwu, line 1187: image

`null` is not a number.

tonytlwu (changes requested):

tonytlwu:
PR lacks explanation @sofiiakvasnevska

js/utils.js
ivandevupp: They are already nested in `data`
image
And we have same condition below on row 1055:
`var aValue = a.data ? a.data[sortField] : a[sortField];`

js/utils.js
ivandevupp: If `aValue` is equal to 0 result of `parseFloat(0, 10)` would be 0 and `Boolean(0)` will return `false` but 0 is a number and should be compared with other numbers greater or less than 0. Otherwise it will be sorted wrong:
image

js/utils.js
ivandevupp: Yes , but this condition works only when `typeof value === 'string'`

js/utils.js
tonytlwu: Thanks for the explanation. Makes sense now.

tonytlwu (approved):

Risk: Low
Affected: LFDs with sorting enabled
Verify: Entries are sorted correctly

Squash and merge: Improve number parsing for record sorting

PR #464, opened 2020-12-17
OD-253 [Feature] Filter value type User profile data

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-253 https://weboo.atlassian.net/jira/software/projects/OD/boards/9?selectedIssue=OD-253
## Description
Added logic for filter value type, user profile data to in order to follow the user profile data fields
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYlb2Z8vgg
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


js/layout-javascript/agenda-code.js
```diff
@@ -856,6 +856,10 @@ DynamicList.prototype.initialize = function() {
         uuid: _this.data.uuid,
         container: _this.$container,
         records: records
+      }).then(function() {
+        return Fliplet.User.getCachedSession();
+      }).then(function(session) {
+        _this.Utils.Records.setSource(session, _this.data);
```
tonytlwu, line 862: `_this.Utils.Records.setSource()` returns a promise, which suggests in most cases that we should wait for the promises to resolve before continuing.

```suggestion
        return _this.Utils.Records.setSource(session, _this.data);
```

js/layout-javascript/agenda-code.js
```diff
@@ -856,6 +856,10 @@ DynamicList.prototype.initialize = function() {
         uuid: _this.data.uuid,
         container: _this.$container,
         records: records
+      }).then(function() {
+        return Fliplet.User.getCachedSession();
```
tonytlwu, line 860: If "User profile data" filter type isn't used at all , then `Fliplet.User.getCachedSession()` doesn't need to be called at all.

js/utils.js
```diff
@@ -1588,6 +1630,7 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
       getFields: getRecordFields,
       getFieldValues: getRecordFieldValues,
       parseFilters: parseRecordFilters,
+      setSource: setSourceValue,
```
tonytlwu, line 1633: Use a function that takes the LFD configuration as input and then based on the configuration, loads all the necessary values.

```suggestion
      setFilterValues: setFilterValues,
```

tonytlwu (changes requested):

AndrRyaz:
@tonytlwu thank you for your notes, I added fixes and will take into account in the following features

js/layout-javascript/news-feed-code.js
```diff
@@ -1112,9 +1112,7 @@ DynamicList.prototype.initialize = function() {
         container: _this.$container,
         records: records
       }).then(function() {
-        return Fliplet.User.getCachedSession();
-      }).then(function(session) {
-        _this.Utils.Records.setSource(session, _this.data);
+        return _this.Utils.Records.setFilterValues(_this.data);
```
tonytlwu, line 1115: Use an embedded options object so that we can add more options to `_this.Utils.Records.setFilterValues()` if necessary without having to alter `_this.data`.

```suggestion
        return _this.Utils.Records.setFilterValues({
          config: _this.data
        });
```

js/utils.js
```diff
@@ -1511,38 +1511,41 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
     });
   }
 
-  function setSourceValue(options, data) {
+  function setFilterValues(data) {
```
tonytlwu, line 1514:
```suggestion
  function setFilterValues(options) {
```

js/utils.js
```diff
@@ -1511,6 +1511,51 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
     });
   }
 
+  function setFilterValues(data) {
+    return new Promise(function(resolve) {
+      if (!data.filterOptions.length) {
```
tonytlwu, line 1516: In case the filter options don't exist.

```suggestion
      if (!(data.filterOptions || []).length) {
```

js/utils.js
```diff
@@ -1511,6 +1511,51 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+        switch (item.valueType) {
+          case 'user-profile-data':
+            Fliplet.User.getCachedSession()
+              .then(function(options) {
+                var entries = options.entries;
+
+                if (options && options.entries) {
+                  if (entries.dataSource && entries.dataSource.data.hasOwnProperty(item.fieldValue)) {
+                    item.value = entries.dataSource.data[item.fieldValue];
+                    return resolve();
```
tonytlwu, line 1530: If there are multiple filter options, this one single `resolve()` on the first loop will resolve the `setFilterValues()` promise. This won't work.

tonytlwu (changes requested):

js/utils.js
```diff
@@ -1511,47 +1511,58 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
     });
   }
 
-  function setFilterValues(data) {
+  function setFilterValues(options) {
     return new Promise(function(resolve) {
-      if (!data.filterOptions.length) {
+      if (!(options.config.filterOptions || []).length) {
```
tonytlwu, line 1516: Add `options = options || {}` at the top. There needs to be more error checking.

And check if `options.config` exists too.

js/utils.js
```diff
@@ -1511,47 +1511,58 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
-                }
-
-                Fliplet.Profile.get(item.fieldValue)
-                  .then(function(result) {
-                    item.value = result || '';
-                    return resolve();
+      var promises = [];
+
+      options.config.filterOptions.forEach(function(item, index) {
+        promises.push(
```
tonytlwu, line 1523: Can we reduce one level of indentation with:

```js
var promise = new Promise(...);

promises.push(promise);
```

js/utils.js
```diff
@@ -1511,47 +1511,58 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+          new Promise(function(res) {
+            switch (item.valueType) {
+              case 'user-profile-data':
+                Fliplet.User.getCachedSession()
```
tonytlwu, line 1527: If there are 10 rules all using profile data, then `Fliplet.User.getCachedSession()` is being called 10 times. The result is the same so it really only need to be called once. On native apps calling this requires accessing native storage data and so 10 calls would create unnecessary lag.

js/utils.js
```diff
@@ -1511,47 +1511,58 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
-
-                Fliplet.Profile.get(item.fieldValue)
-                  .then(function(result) {
-                    item.value = result || '';
-                    return resolve();
+      var promises = [];
+
+      options.config.filterOptions.forEach(function(item, index) {
+        promises.push(
+          new Promise(function(res) {
```
tonytlwu, line 1524:
```suggestion
          new Promise(function(resolve) {
```

js/utils.js
```diff
@@ -1511,6 +1511,62 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
     });
   }
 
+  function setFilterValues(options) {
+    return new Promise(function(resolve) {
```
tonytlwu, line 1515: You don't need to start a `new Promise()` here.

At the bottom of the function you could just use `return Promise.all(promises)`

tonytlwu (changes requested):
See comments. The approach isn't appropriate.

js/utils.js
AndrRyaz: @tonytlwu,
as I understand it, the new Promise wrapper is needed for the `getCachedSession` function because we are waiting for a callback from the function, if we remove the wrapper, then the main function is not executed due to the fact that the callback does not have time to complete and the function occurs asynchronously.
Here is a vide without `new Promise` wrapper
http://somup.com/cYl32H2HSV

js/utils.js
tonytlwu: ...or use:

```js
var promises = _.map(options.config.filterOptions, fn);
```

js/utils.js
tonytlwu: @AndrRyaz I disagree. Tell me if the following structure misses any important race conditions.

```js
function setFilterValues(options) {
  options = options || {};

  if (!options.config) {
    return Promise.resolve();
  }

  // Using lodash function so the IF clause above doesn't need to check for options.config.filterOptions
  var promises = _.map(options.config.filterOptions, function() {
    return new Promise(function(resolve) {
      switch () {
        // Simplified summary of your current code (excl. other recommended changes)
        case 'use-profile-values':
          Fliplet.User.getCachedSession()
            .then(function(session) {
              item.value = data.entries.dataSource.data[item.fieldValue];
              resolve();
            });
          break;
        // ... 
      }
    });
  });

  // setFilterValues returns a promise that doesn't resolve until all promises have resolved
  return Promise.all(promises);
}
```

AndrRyaz:
@tonytlwu thanks for your explanation, I'll take this point to my mind.

js/utils.js
```diff
@@ -1512,62 +1512,30 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
   }
 
   function setFilterValues(options) {
+    // Set options default in case it's undefined
```
tonytlwu, line 1515: We don't need this comment. We do this a lot. It looks like you've included comments from my code examples. Those comments were to walk you through the code example. We don't need all of them.

```suggestion
```

AndrRyaz: @tonytlwu,  removed the comments

js/utils.js
```diff
@@ -1511,6 +1511,33 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+    }
+
+    return new Promise(function(resolve) {
+      Fliplet.User.getCachedSession()
+        .then(function(session) {
+          if (options.config.filterOptions.length) {
+            return Promise.all(_.map(options.config.filterOptions, function(item) {
+              switch (item.valueType) {
+                case 'user-profile-data':
+                  item.value = session.entries ? session.entries.dataSource.data[item.fieldValue] : '';
```
tonytlwu, line 1528: Your previous version of the code checked for all other passport too. I don't understand why that was removed. I provided some [code sample](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/464#discussion_r547386465) for structural purpose. It looks like there hasn't been any thought in including the good work that was previously done.

js/utils.js
```diff
@@ -1511,6 +1511,33 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
 
+  function setFilterValues(options) {
+    options = options || {};
+
+    if (!options.config) {
+      return Promise.resolve();
+    }
+
+    return new Promise(function(resolve) {
+      Fliplet.User.getCachedSession()
```
tonytlwu, line 1522: If there's no filter option with the value type `user-profile-data` then you don't need to get the user's cached session.

You should have some logic that checks through the filter options first to determine whether user session is required. This logic will end up getting expanded to check for app storage data as well when we add support for app storage filter vlaues.

tonytlwu (changes requested):

js/utils.js
AndrRyaz: @tonytlwu 
This is not true, your example code was very useful, but why did I think that the version was incomplete after deleting comments. i went back to the previous version of the code, and got a little confused about the implementation sorry

js/utils.js
AndrRyaz: @tonytlwu 
 I used your example code https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/464#discussion_r547386465 in which `getCachedSession` will also be called every time if the case is `user-profile-data`, but I also remember that you asked that this function would not be called several times that in the end I made in this commit 
https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/464/commits/c2078e6bfa5b6716b87864548d94101e5dd79374
And what would be better to call once and use in the right case or call every time in the case

tonytlwu: It should only be called once, and only when it's necessary. The code would look through all filter options to find out if user profile data is needed. If so, it calls `getCachedSession` once to get the session data before looping through the filter options and applying the values. And, when applying the values, all passport entry types should be considered as you had originally done.

It seems that your several commits and reverts seem to be missing one or more of these points. We need a solution that meets all the requirements in the ticket while performing at its best.

AndrRyaz:
@tonytlwu i updated the logic as required

tonytlwu (dismissed):

Risk: High
Affected: LFDs with filters configured in Studo
Verify: All existing filters work as configured

Squash and merge: Adds support for filtering with user profile data

PR #466, opened 2020-12-17
OD-295 [Fix] Make Data source not disappearing after failed form validation

ivandevupp:
@sofiiakvasnevska
## Issue
OD-295 https://weboo.atlassian.net/browse/OD-295
## Description
Make Data source not disappearing after failed form validation with required fields
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYlb2w8vpE
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -110,12 +110,16 @@ var DynamicLists = (function() {
       selector: '#dataSourceProvider',
       data: dataSourceData,
       onEvent: function(event, dataSource) {
-        if (event === 'dataSourceSelect' && dataSource.columns) {
-          dataSourceColumns = dataSource.columns;
-          _this.updateSummaryRowContainer();
-          _this.updateDetailsRowContainer();
-          _this.renderSortColumns();
-          _this.renderFilterColumns();
+        if (event === 'dataSourceSelect') {
+          _this.config.dataSourceId = dataSource.id;
+
+          if (dataSource.columns) {
+            dataSourceColumns = dataSource.columns;
+            _this.updateSummaryRowContainer();
+            _this.updateDetailsRowContainer();
+            _this.renderSortColumns();
+            _this.renderFilterColumns();
+          }
         }
       }
     });
--- a/js/interface.js
+++ b/js/interface.js
@@ -124,14 +124,8 @@ var DynamicLists = (function() {
       }
     });
 
-    dataSourceProvider.then(function(dataSource) {
-      _this.config.dataSourceId = dataSource.data.id;
-    });
+    _this.config.dataSourceProvider = dataSourceProvider;
   }
-
-  Fliplet.Widget.onSaveRequest(function() {
-    dataSourceProvider.forwardSaveRequest();
-  });
 
   DynamicLists.prototype = {
     // Public functions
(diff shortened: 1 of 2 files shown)


js/interface.js
```diff
@@ -124,14 +124,10 @@ var DynamicLists = (function() {
       }
     });
 
-    dataSourceProvider.then(function(dataSource) {
-      _this.config.dataSourceId = dataSource.data.id;
-    });
-  }
+    _this.config.dataSourceProvider = dataSourceProvider;
```
tonytlwu, line 127: I don't like we're adding to the `config` object content that isn't technically the widget configuration. It's being used here as a way to provide access to the provider promise in another file. There are many other ways to do this without unnecessarily changing the widget configuration.

Also, rather than triggering an event for an event listener to execute the intialisation, I'd rather we define a function to be executed.

See https://developers.fliplet.com/API/core/registry.html for how to use `Fliplet.Registry` to achieve this.

Examples:

- https://github.com/Fliplet/fliplet-widget-data-source/blob/487cb5f37007e3e3cc7d46f1e28edb4b26fc3623/js/preconfigured-rules.js
- https://github.com/Fliplet/fliplet-widget-notification-inbox/blob/8b44f8173578c95fc310f7e042b64dd7608ee5a9/js/app/libs.js
- https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/186709583e62c430c25d3fe8c369872dc59cf339/js/query-parser.js

tonytlwu (changes requested):
PR lacks explanation

js/interface.js
ivandevupp: I added DSP by using `Fliplet.Registry` but I still need to trigger custom event. This event is not for executing initialization, it is used to notificate `interface-list` when to apply `then` for correct DSP saving. 
Otherwise we will need to wrap whole initialization in function to be executed, but this may cause more problems and take too much time.

js/interface.js
```diff
@@ -84,6 +84,10 @@ var DynamicLists = (function() {
 
     _this.attachListeners();
     _this.init();
+
+    Fliplet.Registry.set('datasource-provider', function() {
```
tonytlwu, line 88: This will do.

```suggestion
    Fliplet.Registry.set('datasource-provider', dataSourceProvider);
```

tonytlwu (changes requested):

js/interface.js
ivandevupp: I tried setting it without function but it didn't work because before `dataSourceProvider` initialization it is `undefined`.
Otherwise if I set it after initialization I can't get in on init in `interface-list` because it is still not set.
Setting it as callback let us resolve this value when it is set and use `Registry` on inits.
That's why I created event that triggers when DSP is initialized in `interface` and then I can perform needed actions in `interface-list`.

tonytlwu: @ivandevupp Thanks for explaining. That makes sense now 👍🏻

tonytlwu (approved):

Risk: High
Verify: Data source selection and configurations are correctly loaded & saved

Squash and merge: Keep data source provider visible after failed form validation

PR #467, opened 2020-12-17
OD-297 [Fix] Overlay filter's height

ivandevupp:
@sofiiakvasnevska
## Issue
OD-297 https://weboo.atlassian.net/browse/OD-297
## Description
Filter overlay now is not hidden when the LFD is under an Accordion.
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYlbo48vWd
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/css/layout-css/small-card-style.upload.css
+++ b/css/layout-css/small-card-style.upload.css
@@ -12,7 +12,6 @@
   top: 0;
   top: env(safe-area-inset-top);
   right: 0;
-  bottom: 0;
   left: 0;
   z-index: 11;
   background-color: transparent;
--- a/css/layout-css/small-card-style.upload.css
+++ b/css/layout-css/small-card-style.upload.css
@@ -18,6 +18,8 @@
   background-color: transparent;
   transform: translate3d(100%, 0, 0);
   transition: all 0.2s ease-out;
+  height: 100vh;
+  padding-bottom: env(safe-area-inset-top);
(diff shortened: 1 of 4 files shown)


tonytlwu (changes requested):
Lacks explanation on how this works

ivandevupp:
When LFD is added in Accordion it narrows filter in overlay because it has auto height, now its height is set to viewport height with padding bottom of  `safe-area-inset-top` to show all items when they are expanded. These changes affect only on overlay filter.

css/layout-css/small-card-style.upload.css
```diff
@@ -18,6 +18,8 @@
   background-color: transparent;
   transform: translate3d(100%, 0, 0);
   transition: all 0.2s ease-out;
+  height: 100vh;
```
tonytlwu, line 21: If the height is now statically set to `100vh` with a padding at the bottom, then does this mean `bottom: 0` is no longer necessary? If so, please remove it.

ivandevupp: Yes, thank you, I removed it.

tonytlwu (approved):

Risk: Medium
Affected: LFD users with filter overlays on
Verify: Filter overlays look correct

Squash and merge: Set a height for filter overlays

PR #468, opened 2020-12-18
OD-286 [Fix] 'Field where the post author's email is stored' in LFD entry management will be updated as expected

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-286

## Description
Added update of the admin's email field selection on the data source change

## Screenshots/screencasts
https://share.getcloudapp.com/2Nu0z7Je

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/interface.js
```diff
@@ -119,6 +119,7 @@ var DynamicLists = (function() {
             _this.updateDetailsRowContainer();
             _this.renderSortColumns();
             _this.renderFilterColumns();
+            _this.updateFieldsWithColumns(dataSource.columns);
```
tonytlwu, line 122: Generally it's better once a variable is assigned to another variable, to use the new variable -- unless something else might be changing it before it's used.

Do any of the 4 lines above change `dataSourceColumns`? If not, then see below.

```suggestion
            _this.updateFieldsWithColumns(dataSourceColumns);
```

tonytlwu (changes requested):
Lacks clear explanation on how this works

js/interface.js
YaroslavOvdii: @tonytlwu It seems that `dataSourceColumns` isn't changing in any of those functions. I've change the code as you've suggested.

tonytlwu:
@sofiiakvasnevska I mentioned last week that this lacks explanation on how the fix works. It's still lacking.

YaroslavOvdii:
@tonytlwu we are reusing an existing function that was used for filling admins fields select before.
We put it on the event where we listening when the data source is changed so when we select another data source our admins fields select will be filled with correct columns.

tonytlwu (approved):

Risk: Medium
Verify: Users can switch data sources without losing configurations unnecessarily

Squash and merge: Update fields using new data source selection

PR #470, opened 2020-12-21
OD-294 [Fix] Show selected None and Custom security rule after clicking 'Back to Settings'

ivandevupp:
@sofiiakvasnevska
## Issue
OD-294 https://weboo.atlassian.net/browse/OD-294
## Description
 Show selected None and Custom security rule after clicking 'Back to Settings'
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYlr6a8pFJ
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -110,12 +110,16 @@ var DynamicLists = (function() {
       selector: '#dataSourceProvider',
       data: dataSourceData,
       onEvent: function(event, dataSource) {
-        if (event === 'dataSourceSelect' && dataSource.columns) {
-          dataSourceColumns = dataSource.columns;
-          _this.updateSummaryRowContainer();
-          _this.updateDetailsRowContainer();
-          _this.renderSortColumns();
-          _this.renderFilterColumns();
+        if (event === 'dataSourceSelect') {
+          _this.config.dataSourceId = dataSource.id;
+
+          if (dataSource.columns) {
+            dataSourceColumns = dataSource.columns;
+            _this.updateSummaryRowContainer();
+            _this.updateDetailsRowContainer();
+            _this.renderSortColumns();
+            _this.renderFilterColumns();
+          }
         }
       }
     });
--- a/js/interface.js
+++ b/js/interface.js
@@ -1201,7 +1201,7 @@ var DynamicLists = (function() {
         }
 
         item.columns = dataSourceColumns || _this.config.defaultColumns;
-        item.column = item.columns.indexOf(item.column) !== -1 ? item.column : null;
+        item.column = _this.getRowColumnValue(item);
         item = _this.updateWithFoldersInfo(item, 'summary');
         _this.addSummaryItem(item);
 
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -1243,6 +1243,13 @@ var DynamicLists = (function() {
         }
       });
     },
+    getRowColumnValue: function(item) {
```
tonytlwu, line 1246: I don't know what `getRowColumnValue` means.

tonytlwu (changes requested):

js/interface.js
ivandevupp: I added description comment, this name is based on function usages

ivandevupp: Would `getSelectedColumnValue` be better for comprehension?

tonytlwu: It sounds like the scenario is that you have an `item`, which contains `item.column`. You're validating `item.column` to make sure it's part of `item.columns`.

Why are `item.column === 'empty'` and `item.column === 'custom'` exceptions?

ivandevupp: Yes, because when we set field value with `empty` or `custom` this value is written in `item.column` and `item.columns` won't include it.

tonytlwu: Why and where is `empty` and `custom` saved into `item.columns`?

ivandevupp: They are saved in `'item.column`, `item.columns` is an array of data source columns or default columns.
When we PUT widget-instance, the `summary-fields` looks like:
image

When we render `Data view options` we have next scenarios for setting selected summary fields' values:

1. Data source or default columns include `item.column` -> show column name
2. Data source or default column doesn't include `item.column` -> show `-- Select a data field`
3. `item.column` is `none` -> show `-- Select a data field`
4. `item.column` is `custom` and `customFieldEnabled === true` -> show `Custom` and custom field
5. `item.column` is `empty` -> show 'None'

Same scenarios are applied to detail-fields
This function `getRowColumnValue` returns `item.column` or `null` to show correct option value in select
At rows 1208 and 1230 we set selected field value as `item.column || 'none'`
If you approve I will change return of this function from `null` to `'none'` and at rows 1208 and 1230 change `item.column || 'none'` to `item.column`.

js/interface.js
```diff
@@ -1223,7 +1223,7 @@ var DynamicLists = (function() {
       $detailsRowContainer.empty();
       _.forEach(_this.config.detailViewOptions, function(item) {
         item.columns = dataSourceColumns;
-        item.column = item.columns.indexOf(item.column) !== -1 ? item.column : null;
+        item.column = _this.getRowColumnValue(item);
```
tonytlwu, line 1226:
```suggestion
        item.column = _this.validateColumn({
          column: item.column,
          columns: item.columns
        });
```

js/interface.js
tonytlwu: @ivandevupp Thanks for explaining. I get it now. In this case, I'd name it slightly different with a different purpose.

```suggestion
    /**
     * Validates a column name from field settings
     */
    validateColumn: function(options) {
```

tonytlwu (changes requested):

squallstar (approved):

- **Risk**: Low
- **Affected**: Studio users configuring the component
- **Verify**: Verify the issue is resolved

PR #471, opened 2020-12-24
OD-303 OD-301 [Fix] Data Source Provider will be adding/enabling all needed access rules for the app

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-303
https://weboo.atlassian.net/browse/OD-301

## Description
This issue occurs when we were trying to add/enabling all access rules for the app and in the `for ... in` loop we have mutated the default rule type, with the next value (e.g. it was the `update` but then we added the `delete` rule and previous rule type was also changed to `delete`).
That is why put the declaration of the `defaultRule` into the loop so it wouldn't mutate the previous type value.

## Screenshots/screencasts
https://share.getcloudapp.com/z8u4lymb

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


squallstar (approved):

squallstar:

- **Risk**: Low
- **Affected**: Components UIs with data source selection
- **Verify**: Verify the interface can be used to select the data source

PR #473, opened 2020-12-29
OD-258 [Feature] Filter logic: Number is between

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-258 https://weboo.atlassian.net/jira/software/projects/OD/boards/9?selectedIssue=OD-258
## Description
Added new logic for `Filter logic: Number is between`, which are filtering the list between two values
## Screenshots/screencasts
http://somup.com/cYlU6r2MTV
## Backward compatibility
## Reviewers


js/interface.js
```diff
@@ -236,19 +236,27 @@ var DynamicLists = (function() {
...
-            $selector.find('.panel-title-text .column').html(value === 'none' ? '(Field)' : value);
-          }
+          switch (type) {
+            case 'field':
+              $selector.find('.panel-title-text .column').html(value === 'none' ? '(Field)' : value);
+              break;
 
-          if (type === 'logic') {
-            var hideValueFields = value === 'empty' || value === 'notempty';
+            case 'logic':
```
squallstar, line 244: what does "case logic" mean?

js/interface.js
```diff
@@ -236,19 +236,27 @@ var DynamicLists = (function() {
...
-          }
+              $selector.find('.panel-title-text .value, #value-dash, #filter-value-type').toggleClass('hidden', hideValueFields);
+              $selector.find('.panel-title-text .value, #value-dash, #filter-value').toggleClass('hidden', hideValueFields);
 
-          if (type === 'valueType') {
-            $selector.find('#filter-value label').html(value !== 'enter-value' ? 'Value for' : 'Value');
+              $selector.find('.panel-title-text .value, #value-dash, #logic-comparasion').toggleClass('hidden', !isLogicComparasion);
+              break;
+
+            case 'valueType':
```
squallstar, line 254: what is a "valueType"?

js/interface.js
```diff
@@ -1906,10 +1922,14 @@ var DynamicLists = (function() {
 
       $filterAccordionContainer.append($newPanel);
 
-      if (data.logic === 'empty' || data.logic === 'notempty') {
+      if (data.logic === 'empty' || data.logic === 'notempty' || data.logic === 'between') {
```
squallstar, line 1925: let's use indexOf?

```js
['a', 'b', 'c'].indexOf(foo) !== -1
```

squallstar (commented):
@tonytlwu can you review this too?

js/interface.js
AndrRyaz: @squallstar 
This logic works when changing any dropdown in the filter list, where type is responsible for which dropdown was changed.
изображение

js/interface.js
AndrRyaz: @squallstar sure, already fixed

js/interface.js
squallstar: @tonytlwu can you confirm the screenshot above is in line with your expectations both in terms of UI and naming of the variables?

It's a little confusing to me (without knowing the feature I would have expected `type` to have values from a dropdown rather than be hardcoded from labels/underlying logic)

AndrRyaz: @squallstar 
Such an implementation has already been added earlier, I decided to change the structure instead of if conditions, I changed to use a switch case as there was a need to add a new condition

AndrRyaz:
The new view of logic dropdown with a new option `Is between`
изображение

js/interface.js
squallstar: I see, that makes sense thanks for explaining.

js/interface.js
```diff
@@ -738,8 +746,16 @@ var DynamicLists = (function() {
         _this.addFilterItem(item);
         $('#select-data-field-' + item.id).val(item.column);
         $('#logic-field-' + item.id).val(item.logic);
-        $('#value-type-field-' + item.id).val(item.valueType);
-        $('#value-field-' + item.id).val(item.value);
+        if (item.logic !== 'between') {
```
squallstar, line 749: empty line before block conditions

AndrRyaz: @squallstar  fixed

squallstar (changes requested):
You have a typo in variables and templates, please change "comparasion" to "comparison"

AndrRyaz:
@squallstar fixed

js/interface.js
```diff
@@ -236,19 +236,27 @@ var DynamicLists = (function() {
...
-            var hideValueFields = value === 'empty' || value === 'notempty';
+            case 'logic':
+              var hideValueFields = ['empty', 'notempty', 'between'].indexOf(value) !== -1;
+              var isLogicComparison = value === 'between';
 
-            $selector.find('.panel-title-text .value, #value-dash, #filter-value-type').toggleClass('hidden', hideValueFields);
-            $selector.find('.panel-title-text .value, #value-dash, #filter-value').toggleClass('hidden', hideValueFields);
-          }
+              $selector.find('.panel-title-text .value, #value-dash, #filter-value-type').toggleClass('hidden', hideValueFields);
+              $selector.find('.panel-title-text .value, #value-dash, #filter-value').toggleClass('hidden', hideValueFields);
```
tonytlwu, line 249:
```js
$selector.find('.panel-title-text .value, #value-dash, #filter-value-type').toggleClass('hidden', hideValueFields);
$selector.find('.panel-title-text .value, #value-dash, #filter-value').toggleClass('hidden', hideValueFields);
```

Why is the above not just:

```js
$selector.find('.panel-title-text .value, #value-dash, #filter-value-type, #filter-value').toggleClass('hidden', hideValueFields);
```

js/interface.js
```diff
@@ -236,19 +236,27 @@ var DynamicLists = (function() {
...
+
+              $selector.find('.panel-title-text .value, #value-dash, #logic-comparison').toggleClass('hidden', !isLogicComparison);
```
tonytlwu, line 251:
```js
$selector.find('.panel-title-text .value, #value-dash, #logic-comparison').toggleClass('hidden', !isLogicComparison);
```

The `hidden` class for `.panel-title-text .value, #value-dash` is toggled based on `hideValueFields` and then `!isLogicComparison`.

If the class needs to be toggled based on a combination of both logic, e.g. `hideValueFields && !isLogicComparison` then please do this only once. Or, if it's only toggled based on `!isLogicComparison` then remove the selector in the previous lines.

js/interface.js
tonytlwu: I'm good with this.

js/interface.js
```diff
@@ -738,8 +746,17 @@ var DynamicLists = (function() {
         _this.addFilterItem(item);
         $('#select-data-field-' + item.id).val(item.column);
         $('#logic-field-' + item.id).val(item.logic);
-        $('#value-type-field-' + item.id).val(item.valueType);
-        $('#value-field-' + item.id).val(item.value);
+
+        if (item.logic !== 'between') {
+          $('#value-type-field-' + item.id).val(item.valueType);
+          $('#value-field-' + item.id).val(item.value);
+        } else {
```
tonytlwu, line 753: Please use the following structure.

```js
if (foo) {
  // ...
  return;
}

// ...
```

templates/interface/filter-panels.interface.hbs
```diff
@@ -64,26 +65,75 @@
         </div>
       </div>
 
-      <div class="form-group clearfix" id="filter-value-type">
-        <div class="col-sm-4 control-label">
-          <label>Value type</label>
+      <div id="default-comparison">
```
tonytlwu, line 68: This template is repeated for each panel. The HTML template should not contain any IDs unless it's specific to each panel, e.g. `value-type-field-{{id}}`. There are several instances in this template. Please correct them all.

tonytlwu (changes requested):

AndrRyaz:
@tonytlwu your comment's are fixed

js/interface.js
AndrRyaz: when doing the task was based on this structure because it already existed

js/interface.js
AndrRyaz: изображение

when doing the task was based on this structure because it already existed

js/interface.js
```diff
@@ -1946,12 +1948,11 @@ var DynamicLists = (function() {
...
 
       if (['empty', 'notempty', 'between'].indexOf(data.logic) !== -1) {
-        $newPanel.find('.panel-title-text .value, #value-dash, #filter-value-type').addClass('hidden');
-        $newPanel.find('.panel-title-text .value, #value-dash, #filter-value').addClass('hidden');
+        $newPanel.find('.panel-title-text .value, #value-dash, #filter-value-type, #filter-value').addClass('hidden');
       }
 
       if (data.logic !== 'between') {
-        $newPanel.find('.panel-title-text .value, #value-dash, #logic-comparison').addClass('hidden');
+        $newPanel.find('#logic-comparison-' + data.id).addClass('hidden');
```
tonytlwu, line 1955: You don't need the context of `$newPanel` if you're referring to an element of unique ID on the page. Please correct all other similar mistakes in this PR too.

```suggestion
        $('#logic-comparison-' + data.id).addClass('hidden');
```

templates/interface/filter-panels.interface.hbs
```diff
@@ -65,75 +65,73 @@
...
-          </div>
-          <div class="col-sm-8">
-            <label for="value-type-field-{{id}}" class="select-proxy-display">
-              <select name="value-type-field-{{id}}" id="value-type-field-{{id}}" data-field="valueType" class="hidden-select form-control">
-                <option value="enter-value">Enter a value</option>
-              </select>
-              <span class="icon fa fa-chevron-down"></span>
-            </label>
-          </div>
+      <div class="form-group clearfix" id="filter-value-type">
```
tonytlwu: The assumption in my previous review is that this `.hbs` template is repeated for each filter. That means this element ID is not unique to this filter. I mentioned all similar mistakes in this template needs to be fixed. Is this an exception for any good reason? If not, please fix them all in this file.

```suggestion
      <div class="form-group clearfix">
```

tonytlwu (changes requested):

js/interface.js
AndrRyaz: @tonytlwu I corrected everything about your comment in the pr

templates/interface/filter-panels.interface.hbs
AndrRyaz: @tonytlwu 
I added this one identifier to add a "hidden" class to the block, which is  used in other logic options, but during the choice "between logic" I added 2 of the same blocks with changed labels, in my opinion, this is clearer and clearer for further work with logic

tonytlwu: I'm not sure why there's any further discussion required around this.

[The file](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/473/files#diff-6580d9e7b0e71f2c9a22ce6a2103b2bc6aced2c8c25fc51a3271104665c7fd41) is repeated for each filter, therefore there the template should not contain any `id` attribute that isn't unique to each repetition because it will create 2 elements of the same ID -- we can't have this.

AndrRyaz: @tonytlwu 
I added an ID for the required tags.

tonytlwu: There are still elements in the file that has an `id` that gets repeated.

https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/4e2652dd518eab3632968dc1a056717ebd401bbe/templates/interface/filter-panels.interface.hbs#L87

AndrRyaz: @tonytlwu, sorry missed this moment when resolved conflicts

templates/interface/filter-panels.interface.hbs
```diff
@@ -51,21 +51,83 @@
...
           <label for="value-field-{{id}}">Value</label>
         </div>
         <div class="col-sm-8">
           <input type="text" name="value-field-{{id}}" id="value-field-{{id}}" data-field="value" class="form-control">
         </div>
       </div>
+    </div>
+
+    <div id="logic-comparison-{{id}}">
+      <div class="form-group clearfix" id="filter-value-type-from-{{id}}">
```
tonytlwu, line 87: What's this ID used for? It seems unnecessary as I don't see it being referred to in CSS or JS.

```suggestion
      <div class="form-group clearfix">
```

AndrRyaz: I believe that after resolving conflicts, I lost part of the code, I will now double-check the history, sorry

AndrRyaz: @tonytlwu, yes you're right ids is not used, sorry, already removed them

tonytlwu (dismissed):

Risk: High
Affected: LFDs with filters configured in Studio
Verify: LFD filter configuration works as expected

Squash and merge: Add filter logic for "number is between"

PR #479, opened 2020-12-31
OD-309 [Fix] List sorting token fields will be updated on data source change

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-309

## Description
The issue was that we didn't update the token fields of the list sorting.
I've added it and took an example from the list filter that was above.

## Screenshots/screencasts
https://share.getcloudapp.com/geuo1WxW

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/interface.js
```diff
@@ -1352,6 +1352,19 @@ var DynamicLists = (function() {
         showAutocompleteOnFocus: true,
         createTokensOnBlur: false
       });
+      $('.sort-fields').html(tokenField({
```
tonytlwu, line 1355: Similar code is already found about 26 lines above. This looks like unnecessary duplicates.

tonytlwu (changes requested):

js/interface.js
YaroslavOvdii: @tonytlwu Yes, you are right. I'll remove it. The issue was that we were using an old variable at previous lines I've fixed that.

tonytlwu (approved):

Risk: Medium
Affected: Sort UI configuration for LFDs
Verify: Users can configure and save LFD sort fields

Squash and merge: Update sort fields when data source is changed

PR #517, opened 2021-01-26
OD-19 [Feature] Show a list of the files associated to each LFD entry

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-19

## Description
We have added a new detail view for the file field type

## Screenshots/screencasts
https://share.getcloudapp.com/bLugYD5Y

## Backward compatibility

## Reviewers


YaroslavOvdii:
There is one more thing. As you may see in a video we don't have a size value. That is because when we get file data from the `Fliplet.Media.Files.get(id)` it is returning a size `null`  no matter what kind of file I'm uploading.
Is there any way could fix it on our side?

templates/build/small-h-card-detail.build.hbs
```diff
@@ -123,6 +123,30 @@
           \{{/if}}
         \{{/ifCond}}
 
+	      <!-- File template -->
```
tonytlwu, line 126:
```suggestion
        <!-- File template -->
        \{{#ifCond type '==' 'file'}}
```

js/layout-javascript/agenda-code.js
```diff
@@ -756,6 +756,11 @@ DynamicList.prototype.attachObservers = function() {
         Fliplet.UI.Actions(options);
       });
     })
+    .on('click', '.file-item', function(file) {
+      var url = $(file.currentTarget).find('input[type=hidden]').val();
```
tonytlwu, line 760:
```suggestion
      var url = $(file.target).find('input[type=hidden]').val();
```

js/layout-javascript/agenda-code.js
```diff
@@ -2451,58 +2460,80 @@ DynamicList.prototype.showDetails = function(id, listData) {
...
+  })
+    .then(function(files) {
+      entryData = _this.addDetailViewData(entryData);
 
-  var beforeShowDetails = Promise.resolve({
-    src: src,
-    data: entryData
-  });
+      if (files && Array.isArray(files)) {
+        files.forEach(function(file) {
```
tonytlwu, line 2471: You can save yourself the `if` above with this:

```suggestion
       _.forEach(files, function(file) {
```

js/layout-javascript/agenda-code.js
```diff
@@ -2451,58 +2460,80 @@ DynamicList.prototype.showDetails = function(id, listData) {
...
+            return;
+          }
 
-  if (typeof _this.data.beforeShowDetails === 'function') {
-    beforeShowDetails = _this.data.beforeShowDetails({
-      config: _this.data,
-      src: src,
-      data: entryData
-    });
+          var isFileAdded = entryData.entryDetails.some(function(entryDetail) {
```
tonytlwu, line 2476:
```suggestion
          var isFileAdded = !!_.find(entryData.entryDetails, { id: file.id });
```

js/utils.js
```diff
@@ -45,6 +45,71 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+    var entry = options.entryData;
+    var detailViewFileOptions = _.filter(options.detailViewOptions, function(option) {
+      return option.editable && option.type === 'file';
+    });
+
+    var formFilesInfoInDetailViewOptions = detailViewFileOptions.map(function(detailViewFileOption) {
+      return new Promise(function(resolve) {
+        var label = '';
+        var labelEnabled = true;
+        var files = typeof entry.originalData[detailViewFileOption.column] === 'string' ?
```
tonytlwu, line 58: The `?` and `:` go into the beginning of next line, not the end.

```js
foo = bar
  ? 1
  : 2;
```

js/utils.js
```diff
@@ -45,6 +45,71 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
     return parseFloat(value);
   }
 
+  function getFilesInfo(options) {
+    var entry = options.entryData;
+    var detailViewFileOptions = _.filter(options.detailViewOptions, function(option) {
+      return option.editable && option.type === 'file';
```
tonytlwu, line 51: Why does it need to be `editable`? And if not, use `_.filter(x, { type: 'filter' })` instead.

```suggestion
      return option.type === 'file';
```

js/utils.js
```diff
@@ -45,6 +45,71 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+    });
+
+    var formFilesInfoInDetailViewOptions = detailViewFileOptions.map(function(detailViewFileOption) {
+      return new Promise(function(resolve) {
```
tonytlwu, line 55: Make sure errors are caught too.

```suggestion
      return new Promise(function(resolve, reject) {
```

js/utils.js
```diff
@@ -45,6 +45,71 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+          case 'no-label':
+            labelEnabled = false;
+
+            break;
+          default:
+            break;
+        }
+
+        var filesDetailViewInfo = Promise.all(files.map(function(fileUrl) {
+          var fileId = Fliplet.Media.isRemoteUrl(fileUrl)[2];
```
tonytlwu, line 86: It's sematically weird to use `Fliplet.Media.isRemoteUrl()` to get the file ID. I understand it uses regex so this will probably work. @squallstar Do you approve?

js/utils.js
```diff
@@ -45,6 +45,71 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+
+          return Fliplet.Media.Files.get(fileId).then(function(file) {
```
tonytlwu, line 88: @squallstar I think we have something that could create a server load issue here.

In order to print file information on the page, LFD needs to make a request to find out file information. If there are 1000 LFD entries and each entry has 20 files attached, then this will make 20k requests at once.

I think we'll need:

1. An endpoint and JS API to get the file information for multiple files at once
2. Cache the response so that it doesn't get the file information for a file that has been loaded before (this can be done internally via the JS API or by the LFD code)

Thoughts?

tonytlwu (changes requested):

js/utils.js
squallstar: 1. Agreed (around 2h work)
2. Agreed

js/utils.js
squallstar: 1. This will break if `isRemoteUrl` doesn't return an array
2. This will break if we change the capturing groups (or their order) in `isRemoteUrl`. Technically, this will be more solid if it doesn't rely on it and uses a simple regexp to grab the ID, e.g. `v1\/media\/files/([0-9]+)`

tonytlwu: @squallstar Thanks for your input.

@YaroslavOvdii Please use a regexp to directly grab the file ID instead of using `Fliplet.Media.isRemoteUrl()`

js/utils.js
tonytlwu: @squallstar Will it be possible and suitable for the JS API to handle the caching, and how long will that take?

tonytlwu: This is the mockup for what we're aiming to display in the LFD detail view, based on the data found in each data source entry. The data source column will likely contain a list of file entry IDs for each entry.

image

Each file will need its uploaded date and size displayed. To do so, the LFD needs to get the file metadata. And I assume once this is retrieved, there's really no reason that they would change again, hence the caching recommendation.

js/utils.js
```diff
@@ -45,6 +45,71 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+            return {
+              name: file.name,
+              size: file.size,
```
tonytlwu, line 91: @YaroslavOvdii We need to show the file size, not, in the case of images, the image size.

image

The endpoint response currently seems to be returning the image size, not file size. Can you confirm what this is displaying at the moment?

YaroslavOvdii: @tonytlwu Actually we are receiving a different answer from the API.
In our case, size is always `null`.
We are using `Fliplet.Media.Files.get(fileId)` to get file info.
image

tonytlwu: @YaroslavOvdii I believe that refers to image size, which means the `null` response implies that image size isn't always available. We also need file size information, which appears to be missing.

@squallstar How much time do we need to add file size to the endpoint response? Note that [this thread](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/517#discussion_r564619192) suggests that need the information in bulk.

squallstar: @tonytlwu we already return that whenever available under `metadata.size` (in bytes) but it's only available for files uploaded via the media endpoints (e.g. from the file manager).

tonytlwu: @squallstar Excellent. Then it's just a matter of whether it exists or not.

@YaroslavOvdii See @squallstar's comment above that you need to use `metadata.size` for the file size.

js/layout-javascript/agenda-code.js
YaroslavOvdii: @tonytlwu in our case it would be better to leabe with current traget whitch always will be `li.file-item`. If we use `target` then click may go on child items of the .file-item and our next operation `$(file.target).find('input[type=hidden]').val();` will endup undefined whitch will be leading to the error in this function `Fliplet.Navigate.file(url);`

YaroslavOvdii:
@tonytlwu added suggestions.

Regarding this [task ](https://weboo.atlassian.net/browse/OD-352) in order to do it right, we need to add changes to the appearance component. And it will be done in another PR.

js/layout-javascript/agenda-code.js
tonytlwu: I see. Let's use `event.target` instead of `file.target` then. The parameter to the event listener should always be `event`. Using `file` could be misleading.

tonytlwu:
@YaroslavOvdii You said you added suggestions but I don't see any actual code changes since 26 Jan.

YaroslavOvdii:
@tonytlwu Sorry about that. I've pushed changes in the wrong branch before that. Now added changes correctly.

js/utils.js
```diff
@@ -83,16 +81,19 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
             return {
               name: file.name,
-              size: file.size,
+              size: file.metadata.size,
               uploaded: file.createdAt,
               url: file.url
             };
-          });
+          })
+            .catch(function(err) {
```
tonytlwu, line 94: We can reject the error without parsing it. The parsing should only happen right before something needs to be done with the error, e.g. before displaying it. This way the error object can be passed to the callback function in its full form for assessment.

```suggestion
            .catch(reject);
```

js/utils.js
tonytlwu: @squallstar Can you confirm that you've done both of the below as previously mentioned:

1. An endpoint and JS API to get the file information for multiple files at once
2. Cache the response so that it doesn't get the file information for a file that has been loaded before (this can be done internally via the JS API or by the LFD code). Did you update the JS API to handle this?

js/utils.js
```diff
@@ -111,12 +112,53 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
   }
 
   function registerHandlebarsHelpers() {
-    Handlebars.registerHelper('formatDate', function(date) {
-      if (!date) {
-        return;
+    Handlebars.registerHelper('plaintext', function(context) {
```
tonytlwu, line 115: `plaintext` and `removeSpaces` should not be added. They were removed in https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/534 because they are now declared in our core libraries.

tonytlwu (changes requested):

templates/build/small-h-card-detail.build.hbs
tonytlwu: Not addressed

js/layout-javascript/agenda-code.js
tonytlwu: Not addressed

squallstar:
@tonytlwu 
1. https://github.com/Fliplet/fliplet-api/pull/4484
2. Caching is handled by the backend, not the JS API. This is more effective and introduces less risks, see linked PR for more reference.

tonytlwu:
@squallstar What risks are introduced with frontend-based caching? The original requirement mentioned specifically us needing frontend-based caching because we don't want a slow connection to become a reason where LFD entry details are slow to render when the device has already retrieved the metadata for specific file(s).

squallstar:
@tonytlwu added in https://github.com/Fliplet/fliplet-api/pull/4543

YaroslavOvdii:
@tonytlwu added suggestions

templates/build/small-h-card-detail.build.hbs
tonytlwu: Not addressed

js/utils.js
```diff
@@ -45,13 +45,110 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+            break;
+          case 'no-label':
+            labelEnabled = false;
+
+            break;
+          default:
+            break;
+        }
+
+        var filesID = files.map(function(fileUrl) {
```
tonytlwu, line 83:
```suggestion
        var fileIDs = files.map(function(fileUrl) {
```

tonytlwu (changes requested):

templates/build/small-h-card-detail.build.hbs
tonytlwu: Not addressed

js/utils.js
tonytlwu: Not addressed

templates/build/small-h-card-detail.build.hbs
YaroslavOvdii: @tonytlwu I don't know how this happened but there are no extra spaces in my code editor to remove, should can I do that by commenting on your suggestion?
image

js/utils.js
YaroslavOvdii: @tonytlwu Fixed

templates/build/small-h-card-detail.build.hbs
YaroslavOvdii: @tonytlwu Fixed

squallstar:
@YaroslavOvdii make sure your editor is using spaces, not tabs and the file is converted to use two spaces indentation.

To confirm, when you select all text in the file you should see the gray dots as seen here in my example:

image

YaroslavOvdii:
@squallstar Yes, I'm using spaces in my code editor.
image

squallstar:
Great, thanks for confirming!

css/build.css
```diff
@@ -167,6 +167,35 @@
...
+}
+
+.file-holder .file-icon {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  position: absolute;
+  right: 0;
+  top: 50%;
+  transform: translate(0, -50%);font-size: 34px;
```
tonytlwu, line 194:
```suggestion
  transform: translate(0, -50%);
  font-size: 34px;
```

tonytlwu (changes requested):
See https://github.com/Fliplet/fliplet-theme-default/pull/158#discussion_r594421631 which suggests more CSS are to be added here instead of in the theme.

css/build.css
YaroslavOvdii: @tonytlwu Fixed

tonytlwu (dismissed):

Risk: High
Affected: LFD detail views
Verify: All types of data are rendered correctly in detail view

YaroslavOvdii:
@tonytlwu resolved conflicts.

tonytlwu (approved):

PR #518, opened 2021-01-26
OD-259 [Feature] Add support for date comparison filters

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-259 https://weboo.atlassian.net/browse/OD-259

## Description

- Merged all filter logic's correctly
- Added new optgroup for logic options
- Added new logics : 'Date is', 'Date is before' 'Date is after' 'Date is between'
- Added a filtering data between 2 dates
- Added an offset for current date & time

## Screenshots/screencasts
Filter logic: Date
http://somup.com/crVOlmogQq
Filter logic: Date is between
http://somup.com/crVOIcog6N
Sorry my recorder cannot record more than 5 min videos.
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko @ivandevupp @YaroslavOvdii


js/interface.js
```diff
@@ -231,19 +240,47 @@ var DynamicLists = (function() {
...
+              break;
+
+            case 'date-to':
+              $('#date-number-to-' + id).toggleClass('hidden', isAddNumberFieldDefault);
+              break;
+
             case 'logic':
-              var hideValueFields = ['empty', 'notempty', 'between'].indexOf(value) !== -1;
+              var hideValueFields = logicOptions.indexOf(value) !== -1;
+              var areLogicDate = ['dateis', 'datebefore', 'dateafter'].includes(value);
```
tonytlwu, line 264:
```suggestion
              var isDateLogic = ['dateis', 'datebefore', 'dateafter'].includes(value);
```

js/interface.js
```diff
@@ -231,19 +240,47 @@ var DynamicLists = (function() {
...
+
+              var isAddNumberField = ['today', 'now'].includes($('#date-field-' + id).val());
+              var isAddNumberFieldFrom = ['today', 'now'].includes($('#date-field-from-' + id).val());
+              var isAddNumberFieldTo = ['today', 'now'].includes($('#date-field-to-' + id).val());
+
               var isLogicComparison = value === 'between';
```
tonytlwu, line 270:
```suggestion
              var isBetween = value === 'between';
```

js/interface.js
```diff
@@ -231,19 +240,47 @@ var DynamicLists = (function() {
...
+              var isDateBetweenLogic = value === 'datebetween';
```
tonytlwu, line 271:
```suggestion
              var isDateBetween = value === 'datebetween';
```

js/interface.js
```diff
@@ -231,19 +240,47 @@ var DynamicLists = (function() {
...
+            case 'date-to':
+              $('#date-number-to-' + id).toggleClass('hidden', isAddNumberFieldDefault);
+              break;
+
             case 'logic':
-              var hideValueFields = ['empty', 'notempty', 'between'].indexOf(value) !== -1;
+              var hideValueFields = logicOptions.indexOf(value) !== -1;
+              var areLogicDate = ['dateis', 'datebefore', 'dateafter'].includes(value);
+
+              var isAddNumberField = ['today', 'now'].includes($('#date-field-' + id).val());
```
tonytlwu, line 266:
```suggestion
              var canAddNumber = ['today', 'now'].includes($('#date-field-' + id).val());
```

js/interface.js
```diff
@@ -231,19 +240,47 @@ var DynamicLists = (function() {
...
+              var isAddNumberFieldFrom = ['today', 'now'].includes($('#date-field-from-' + id).val());
```
tonytlwu, line 267:
```suggestion
              var canAddNumberFrom = ['today', 'now'].includes($('#date-field-from-' + id).val());
```

js/interface.js
```diff
@@ -743,21 +776,32 @@ var DynamicLists = (function() {
...
+          $('#number-field-' + item.id).val(item.modifierValue.default);
           $('#value-type-field-' + item.id).val(item.valueType);
+          $('#value-field-' + item.id).val(item.value);
 
           return;
         }
 
-        $('#value-type-field-from-' + item.id).val(item.valueType.from);
-        $('#value-type-field-to-' + item.id).val(item.valueType.to);
+        if (item.logic === 'datebetween') {
```
tonytlwu, line 792: Use `else if`s to connect these `if` clauses

js/interface.js
```diff
@@ -23,6 +23,15 @@ var DynamicLists = (function() {
       ]
     }
   ];
+  var logicOptions = [
```
tonytlwu, line 26: I don't understand what `logicOptions` is and why it only has these 6 logic operators when we have a lot more. It needs to be clearer.

templates/interface/filter-panels.interface.hbs
```diff
@@ -141,5 +194,89 @@
...
+              <optgroup label="&mdash;">
+                <option value="nowsubtractminutes">Now - number of minutes</option>
+                <option value="nowsubtracthours">Now - number of hours</option>
+                <option value="todayminusdays">Today - number of days</option>
+                <option value="todayminusmonths">Today - number of months</option>
+                <option value="todayminusyears">Today - number of years</option>
+              </optgroup>
+            </select>
+            <span class="icon fa fa-chevron-down"></span>
+            <span class="select-value-proxy">-- Please wait</span>
```
tonytlwu, line 226: What's this for?

```suggestion
```

templates/interface/filter-panels.interface.hbs
```diff
@@ -141,5 +194,89 @@
...
```
tonytlwu, line 267:
```suggestion
```

templates/interface/filter-panels.interface.hbs
```diff
@@ -58,13 +58,62 @@
...
+                <optgroup label="&mdash;">
+                  <option value="nowsubtractminutes">Now - number of minutes</option>
+                  <option value="nowsubtracthours">Now - number of hours</option>
+                  <option value="todayminusdays">Today - number of days</option>
+                  <option value="todayminusmonths">Today - number of months</option>
+                  <option value="todayminusyears">Today - number of years</option>
+                </optgroup>
+              </select>
+              <span class="icon fa fa-chevron-down"></span>
+              <span class="select-value-proxy">-- Please wait</span>
```
tonytlwu, line 102:
```suggestion
```

templates/interface/filter-panels.interface.hbs
```diff
@@ -58,13 +58,62 @@
...
               </optgroup>
+              <optgroup label="Date">
+                <option value="dateis">Date is</option>
+                <option value="datebefore">Date is before</option>
+                <option value="dateafter">Date is after</option>
+                <option value="datebetween">Date is between</option>
+              </optgroup>
             </select>
             <span class="icon fa fa-chevron-down"></span>
             <span class="select-value-proxy">-- Please wait</span>
```
tonytlwu, line 69:
```suggestion
```

templates/interface/filter-panels.interface.hbs
```diff
@@ -58,13 +58,62 @@
...
       </div>
 
+      <div id="date-logic-{{id}}">
+        <div class="form-group clearfix">
+          <div class="col-sm-4 control-label">
+            <label>Date</label>
+          </div>
+          <div class="col-sm-8">
+            <label for="date-field-{{id}}" class="select-proxy-display">
+              <select name="date-field-{{id}}" id="date-field-{{id}}" data-field="date" filter-item-id="{{id}}" data-label="-- Select an option" class="hidden-select form-control">
```
tonytlwu, line 81: What's the `data-label` for?

```suggestion
              <select name="date-field-{{id}}" id="date-field-{{id}}" data-field="date" filter-item-id="{{id}}" class="hidden-select form-control">
```

js/interface.js
```diff
@@ -2611,11 +2667,33 @@ var DynamicLists = (function() {
...
           item.value = '';
         }
 
+        if (item.logic === 'datebetween') {
+          var valueDateFrom = $('#date-from-' + item.id).val();
+          var valueDateTo = $('#date-to-' + item.id).val();
+          var valueNumberFrom = $('#number-field-from-' + item.id).val();
+          var valueNumberTo = $('#number-field-to-' + item.id).val();
+
+          item.filterModifier.to = valueDateTo;
```
tonytlwu, line 2690: When the logic is `between`, we use the filter `item.value = { from: valueFrom, to: valueTo }`.

I think we can do the same here with `databetween`:

```js
ite.value = {
  from: { value: 'today' }, // If there's no offset required, don't add the offset property
  to: { value: 'todayadddays', offset: 2 } // Parse the offset value into a number as specified
}
```

js/utils.js
```diff
@@ -458,6 +458,130 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
     }
   }
 
+  function getDateModifiedValues(options) {
+    switch (options.filterModifier) {
+      case 'today':
+        return {
+          today: moment().format('DD/MM/YYYY'),
+          formatedDate: moment(options.date).format('DD/MM/YYYY')
```
tonytlwu, line 466: I don't understand what the requirement is to format it into a string, and more important, why this format.

tonytlwu (changes requested):

js/utils.js
AndrRyaz: @tonytlwu.

When searching for how to correctly compare dates using the moment js, I found more than one option for comparing strings, which I was surprised myself, but it works correctly.
The use of formats was necessary in order to correctly compare dates with an ordered time shift.
For example:
the record in the table is equal to 26 April 2021 16:30
our current date together with the time of April 26, 2021 12:35

to use the shift by the hour, I need to disregard the minutes, as I understood because there will be no coincidence, since the minutes differ, based on this, the Date and check check will almost always be false. But I can also be mistaken in that I have gone too deep into the intricacies of checking the time.

tonytlwu: 1. You can use the functions available in moment for comparison. See https://momentjs.com/docs/#/query/
2. If you want to ignore minutes, then set the minutes to 0 rather than ignoring it through string comparison

For example:

```
moment().startOf('hour') // This sets everything that's less than an hour to 0
```

AndrRyaz: @tonytlwu, okay i'll finish all your comments due next review.

AndrRyaz:
@tonytlwu hello, I added new logic and necessary fixes, I also added a new package that helps the moment.js work with timezones, i added him into widget.json file in assets

js/interface.js
tonytlwu: @sofiiakvasnevska This has not been resolved.

js/utils.js
```diff
@@ -460,6 +611,19 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
           return splitByCommas(filter.value).includes(rowData);
```
tonytlwu, line 611: `Array.prototype.includes()` [doesn't worked in IE](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes).

tonytlwu (changes requested):

js/interface-lists.js
```diff
@@ -457,8 +457,20 @@ function attahObservers() {
         if (widgetData.filterOptions.length) {
           var filterError = [];
           var filterFieldValues = [];
+          var logicOptions = [
```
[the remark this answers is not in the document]
AndrRyaz: @tonytlwu hello sorry for this overlooked point, I thought it was solved in a previous PR - https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/519#discussion_r564663657
We discussed the same, but in a different file
Now I have changed to the same name and the logic of this array in this file carries the same logic as in the one I indicated above

AndrRyaz:
@tonytlwu ,fixed

js/interface.js
```diff
@@ -2610,11 +2693,48 @@ var DynamicLists = (function() {
...
         if (item.logic === 'empty' || item.logic === 'notempty') {
-          item.valueType = null;
+          item.valueType = '';
           item.value = '';
         }
 
+        if (item.logic === 'datebetween') {
+          var valueDateFrom = $('#date-from-' + item.id).val();
+          var valueNumberFrom = $('#number-field-from-' + item.id).val();
+          var isTimeZoneFrom = $('#enable-timezone-from-' + item.id).is(':checked');
```
tonytlwu, line 2713:
```suggestion
          var useDeviceTimezoneFrom = $('#enable-timezone-from-' + item.id).prop('checked');
```

js/utils.js
```diff
@@ -415,6 +415,162 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
     }
   }
 
+  function getDateModifiedValues(options) {
+    var isEntryContainsTime = /^([0-1][0-9]|[2][0-3]):([0-5][0-9])$/.test(options.date);
```
tonytlwu, line 419: 1. Set up `timestamp` variable to simplify the code below.
2. Based on your regex, it's not just that the timestamp "contains time", it contains "only" time. So, use a variable name that's more specific to its purpose.

```suggestion
    var timestamp = options.date;
    var timeOnly = /^([0-1][0-9]|[2][0-3]):([0-5][0-9])$/.test(timestamp);
```

js/utils.js
```diff
@@ -415,6 +415,162 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+    var deviceTimeZone = moment.tz.guess();
+    var getDate = options.isDeviceTimeZoneEnabled ? moment.tz.setDefault(deviceTimeZone) : moment.utc;
+
+    switch (options.filterModifier) {
+      case 'today':
+        return {
+          today: moment().startOf('day'),
+          inputDate: isEntryContainsTime
+            ? moment(options.date, 'HH.mm').startOf('day')
```
tonytlwu, line 428: Why have you chosen to use `HH.mm` with `.` and not the standard `:` syntax?

```suggestion
            ? moment(options.date, 'HH:mm').startOf('day')
```

Also, based on the requirements highlight below, if the entry only contains time, it's not meant to be valid, therefore fails the filter check. Why set the input date as `moment().startOf('day')`?

image

js/utils.js
```diff
@@ -415,6 +415,162 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+            : moment(options.date).startOf('day')
+        };
+
+      case 'now':
+        return {
+          today: getDate().startOf('minute'),
+          inputDate: isEntryContainsTime
```
tonytlwu, line 435: I don't see any checks to verify whether a value that's used for time comparison contains actual time.

image

js/utils.js
```diff
@@ -415,6 +415,162 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+      var result = getDateModifiedValues({
+        date: options.date,
+        filterModifier: options.filterModifier.default.value,
+        modifierValue: options.filterModifier.default.offset,
+        isDeviceTimeZoneEnabled: options.filterModifier.default.isDeviceTimeZoneEnabled
+      });
+
+      switch (options.condition) {
+        case 'dateis':
+          return moment(result.inputDate).isSame(result.today);
```
tonytlwu, line 537: It looks like `result.today` is calculated statically, i.e. does not depend on the value of each record, but is being creating from scratch every single time in the `getDateModifiedValues()` function, e.g. in `case 'now'`, this value is `today: getDate().startOf('minute')`. Why does this need to be created every single time instead of using a cached value?

js/utils.js
```diff
@@ -1311,7 +1475,10 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
       return {
         column: option.column,
         condition: option.logic,
-        value: option.value
+        filterModifier: option.filterModifier,
+        valueType: option.valueType,
+        value: option.value,
+        timezone: option.timezone
```
tonytlwu, line 1481: Where does the `option.timezone` value come from?

js/utils.js
```diff
@@ -1311,7 +1475,10 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
       return {
         column: option.column,
         condition: option.logic,
-        value: option.value
+        filterModifier: option.filterModifier,
```
tonytlwu, line 1478: I don't understand the purpose of `filterModifier`, how it's used and why it's necessary.

tonytlwu (changes requested):

js/interface.js
```diff
@@ -2610,11 +2693,48 @@ var DynamicLists = (function() {
...
+          default: {
+            value: $('#date-field-' + item.id).val(),
+            offset: $('#number-field-' + item.id).val() ? $('#number-field-' + item.id).val() : null,
+            isDeviceTimeZoneEnabled: $('#enable-timezone-default-' + item.id).is(':checked')
+          }
+        };
+
         if (item.logic === 'empty' || item.logic === 'notempty') {
-          item.valueType = null;
+          item.valueType = '';
```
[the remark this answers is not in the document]
AndrRyaz: I changed the value to an empty string due to the fact that all values of this type with a false value have an empty string, for the sake of purity in my opinion, this line was distinguished by the fact that it assigned a different value, although it carries the same functionality as all the others. I have tested and this change has no consequences. But I can change the given variable in reverse

In the case of null, it means that we do not know what a value is, it may or may not exist, we simply do not know. 
An empty string means that we know what the value is and that it is nothing.

js/utils.js
AndrRyaz: This variable has many necessary values such as offset, oseTimezone and others. They are required for every filter.

js/utils.js
AndrRyaz: I cannot move this variable to another place because in this case we can have a scenario in between comparison where we will have in From value with using time zone and in To value without using time zone

js/utils.js
AndrRyaz: Fixed

js/utils.js
AndrRyaz: Fixed

js/interface.js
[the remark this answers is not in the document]
AndrRyaz: I changed this value for consistency. I tested everything works correctly

js/utils.js
AndrRyaz: In this place this variable is unnecessary

js/interface.js
```diff
@@ -221,6 +230,7 @@ var DynamicLists = (function() {
           item.value = '';
           item.valueField = 'Value';
           item.valueType = 'enter-value';
+          item.isDeviceTimeZoneEnabled = false;
```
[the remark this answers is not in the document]
AndrRyaz: Fixed

js/interface.js
```diff
@@ -2610,11 +2693,48 @@ var DynamicLists = (function() {
           item.value = item.fieldValue;
         }
 
+
+        item.filterModifier = {
+          default: {
+            value: $('#date-field-' + item.id).val(),
+            offset: $('#number-field-' + item.id).val() ? $('#number-field-' + item.id).val() : null,
+            isDeviceTimeZoneEnabled: $('#enable-timezone-default-' + item.id).is(':checked')
```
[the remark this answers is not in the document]
AndrRyaz: fixed

js/interface.js
```diff
@@ -232,19 +242,59 @@ var DynamicLists = (function() {
           var type = $(this).data('field');
           var $selector = $(this).parents('.filter-panel');
           var id = $(this).attr('filter-item-id');
+          var isAddNumberFieldDefault = ['today', 'now'].indexOf(value) !== -1;
```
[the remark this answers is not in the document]
AndrRyaz: Fixed

js/interface.js
```diff
@@ -2610,11 +2693,48 @@ var DynamicLists = (function() {
...
+
+          var valueDateTo = $('#date-to-' + item.id).val();
+          var valueNumberTo = $('#number-field-to-' + item.id).val();
+          var isTimeZoneTo = $('#enable-timezone-to-' + item.id).is(':checked');
+
+          item.filterModifier = {
+            default: {
+              value: $('#date-field-' + item.id).val(),
+              offset: $('#number-field-' + item.id).val() ?  $('#number-field-' + item.id).val() : null,
+              isDeviceTimeZoneEnabled: $('#enable-timezone-default-' + item.id).is(':checked')
```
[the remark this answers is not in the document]
AndrRyaz: Fixed

js/interface.js
```diff
@@ -2610,11 +2693,48 @@ var DynamicLists = (function() {
...
+          var isTimeZoneFrom = $('#enable-timezone-from-' + item.id).is(':checked');
+
+          var valueDateTo = $('#date-to-' + item.id).val();
+          var valueNumberTo = $('#number-field-to-' + item.id).val();
+          var isTimeZoneTo = $('#enable-timezone-to-' + item.id).is(':checked');
+
+          item.filterModifier = {
+            default: {
+              value: $('#date-field-' + item.id).val(),
+              offset: $('#number-field-' + item.id).val() ?  $('#number-field-' + item.id).val() : null,
```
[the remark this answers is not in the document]
AndrRyaz: Fixed

AndrRyaz:
@tonytlwu Fixed

js/interface.js
```diff
@@ -2610,11 +2696,47 @@ var DynamicLists = (function() {
...
+          var valueDateFrom = $('#date-from-' + item.id).val();
+          var valueNumberFrom = $('#number-field-from-' + item.id).val();
+          var useDeviceTimezoneFrom = $('#enable-timezone-from-' + item.id).prop('checked');
+
+          var valueDateTo = $('#date-to-' + item.id).val();
+          var valueNumberTo = $('#number-field-to-' + item.id).val();
+          var useDeviceTimezoneTo = $('#enable-timezone-to-' + item.id).prop('checked');
+
+          item.filterModifier = {
+            default: {
```
tonytlwu, line 2722: I assume the `datebewteen` logic requires modifiers for `from` and `to`. Why does it need values for the `default` modifier?

If it's not needed, we shouldn't be saving unnecessary data, which could increase the widget data and page size.

js/interface.js
```diff
@@ -2610,11 +2696,47 @@ var DynamicLists = (function() {
...
+        if (item.logic === 'datebetween') {
+          var valueDateFrom = $('#date-from-' + item.id).val();
+          var valueNumberFrom = $('#number-field-from-' + item.id).val();
+          var useDeviceTimezoneFrom = $('#enable-timezone-from-' + item.id).prop('checked');
+
+          var valueDateTo = $('#date-to-' + item.id).val();
+          var valueNumberTo = $('#number-field-to-' + item.id).val();
+          var useDeviceTimezoneTo = $('#enable-timezone-to-' + item.id).prop('checked');
+
+          item.filterModifier = {
```
tonytlwu, line 2721:
```suggestion
          item.dateFilterModifiers = {
```

js/interface.js
```diff
@@ -2610,11 +2696,47 @@ var DynamicLists = (function() {
           item.value = item.fieldValue;
         }
 
+        item.filterModifier = {
```
tonytlwu, line 2699: Why is `filterModifier` set for all filter options when the modifier isn't necessarily always needed?

js/utils.js
```diff
@@ -1311,6 +1478,8 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
       return {
         column: option.column,
         condition: option.logic,
+        filterModifier: option.filterModifier,
```
tonytlwu, line 1481:
```suggestion
        dateFilterModifiers: option.dateFilterModifiers,
```

js/utils.js
tonytlwu: Assuming there is a filter that uses date comparison, this is my breakdown:

1. `isDateMatches()` is getting called as many times as the the number of records
2. `getDateModifiedValues()` is getting called as many times as there are the number of records
3. The `getDateModifiedValues()` runs something like `getDate().add('minute', smartParseFloat(options.modifierValue)).startOf('minute')` as many times as the number of records. However, `getDate().add('minute', smartParseFloat(options.modifierValue)).startOf('minute')` should produce the same output every single time. If the data source has 10,000 entries, we don't need to compute this value 10,000 times.

You need to fix this problem.

I suggest making sure that the value you pass to `.isSame()`, `.isBefore()` etc. only gets calculated only once for each filter.

tonytlwu (changes requested):
Does not meet coding standard.

AndrRyaz:
@tonytlwu hi, added changes

js/interface.js
```diff
@@ -238,13 +240,47 @@ var DynamicLists = (function() {
               $selector.find('.panel-title-text .column').html(value === 'none' ? '(Field)' : value);
               break;
 
+            case 'date':
+              var dateField = $('#date-field-' + id).val();
```
tonytlwu, line 244: https://github.com/airbnb/javascript/blob/master/README.md#jquery--dollar-prefix

js/interface.js
```diff
@@ -238,13 +248,50 @@ var DynamicLists = (function() {
...
+              $('#enable-timezone-to-' + id).parents('.checkbox').toggleClass('hidden', _this.showTimezoneField(id));
+              break;
+
             case 'logic':
-              var hideValueFields = ['empty', 'notempty', 'between'].indexOf(value) !== -1;
-              var isLogicComparison = value === 'between';
+              var hideValueFields = logicOptionsWithoutValues.indexOf(value) !== -1;
+              var isDateLogic  = ['dateis', 'datebefore', 'dateafter'].indexOf(value) !== -1;
+
+              var canAddNumber = ['today', 'now'].indexOf($('#date-field-' + id).val()) !== -1;
```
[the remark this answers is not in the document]
tonytlwu: @AndrRyaz Not all occurences are resolved.

js/utils.js
```diff
@@ -415,6 +416,191 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+    var timestamp = options.date;
+    var timeOnly = /^([0-1][0-9]|[2][0-3]):([0-5][0-9])$/.test(timestamp);
+    var dateOnly = !/[:]/.test(timestamp);
+    var deviceTimeZone = moment.tz.guess();
+    var getDate = options.useDeviceTimezone ? moment.tz.setDefault(deviceTimeZone) : moment.utc;
+
+    switch (options.dateFilterModifiers) {
+      case 'today':
+        return {
+          today: moment().startOf('day'),
```
tonytlwu, line 441: 1. `moment().startOf('day')` should also go through the same caching system that the other date modifiers are using. See https://jsbench.me/q5klic37ss for how poor `moment().startOf('day')` performs.
2. `today` isn't really "today". Instead of the values `today` and `inputDate`, let's use `comparisonDate` and `entryDate` instead. `comparisonDate` is the control value for comparison and `entryDate` represents the value from each entry.
3. `getCurrentDate()` should be `getCachedDate()`. `getCachedDate()` will load the value from cache if it exists, or create the value from scratch and save it into cache before returning the value.

tonytlwu (changes requested):

AndrRyaz:
@tonytlwu hello i added fixes to variable names and when i checked the comparison again i found 1 miss point and added fixes for it

tonytlwu (commented):
@sofiiakvasnevska Please send this to QA to test on development.

tonytlwu (approved):

Risk: Medium
Affected: All LFD users
Verify: Filter settings are working correctly in Studio and apps

PR #519, opened 2021-01-26
OD-339 [Fix] It's impossible to save LFD with an LFD filter option

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-339 https://weboo.atlassian.net/browse/OD-339
## Description
It's impossible to save LFD with an LFD filter option which uses "Is empty" or "Is not empty" filter logics
## Screenshots/screencasts
http://somup.com/crVO2sogZd
## Backward compatibility
## Reviewers


tonytlwu (changes requested):
Does not conform to coding standards. Does not work in IE 11.

AndrRyaz:
@tonytlwu i changed the logic of the condition to `indexOf `

js/interface-lists.js
```diff
@@ -457,8 +457,20 @@ function attahObservers() {
         if (widgetData.filterOptions.length) {
           var filterError = [];
           var filterFieldValues = [];
+          var nonValidatedLogicOptions = [
```
tonytlwu, line 460: The variable name doesn't make sense.

1. It's `invalid`, not `nonValid`.
2. I don't understand why these are invalid. They look like valid logic options to me. This shows that the variable name is inappropriately named.

tonytlwu (changes requested):

js/interface-lists.js
AndrRyaz: @tonytlwu let me tell you what I mean here please, the logic of this check is to exclude checking of the Value type field when it is hidden and inactive, this array is responsible for each type of logic in which this field is hidden

tonytlwu:
```suggestion
          var logicOptionsWithoutValues = [
```

tonytlwu (changes requested):

AndrRyaz:
@tonytlwu, resolved

tonytlwu (approved):

Risk: Medium
Affected: Studio users configuring LFD filter options and conditions
Verify: Filter options can be correctly saved and loaded

Squash and merge: Ignore logic options that don't need values

PR #520, opened 2021-01-27
OD-107 [Feature] Users will be able to see multiple images in the LFD detail view.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-107

## Description
Added a new logic for the multiple images in the detail view.
Know users will be able to see multiple images and interact with them like in the gallery. 

## Screenshots/screencasts
https://share.getcloudapp.com/bLug6ROl

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko @ivandevupp @AndrRyaz


css/build.css
```diff
@@ -145,6 +145,43 @@
...
+  width: 100%;
+  padding-top: 100%;
+  background-size: contain;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+@media screen and (min-width: 640px) {
+  .multiple-images-item {
+    width: calc(33.33% - 10px);
```
tonytlwu, line 176: Why is this 10px when the mobile one is 7.5px?

```suggestion
    width: calc(33.33% - 7.5px);
```

js/layout-javascript/agenda-code.js
```diff
@@ -2401,6 +2409,27 @@ DynamicList.prototype.addDetailViewData = function(entry) {
       content = entry.originalData[dynamicDataObj.column];
     }
 
+    if (dynamicDataObj.type === 'image') {
+      content = entry.originalData[dynamicDataObj.column];
+
+      if (!Array.isArray(content)) {
```
tonytlwu, line 2415: Let's check strings specifically because `.split()` only works with strings.

```suggestion
      if (typeof content === 'string') {
```

js/layout-javascript/agenda-code.js
```diff
@@ -2401,6 +2409,27 @@ DynamicList.prototype.addDetailViewData = function(entry) {
...
+        content = content.split(/\n/);
```
tonytlwu, line 2416: Why did you make the decision to split the string by linebreaks and not commas, or both?

js/layout-javascript/agenda-code.js
```diff
@@ -2401,6 +2409,27 @@ DynamicList.prototype.addDetailViewData = function(entry) {
...
+      if (!_this.imagesData[dynamicDataObj.id]) {
+        _this.imagesData[dynamicDataObj.id] = {
+          images: [],
+          options: {
+            index: null
+          }
+        };
+      }
+
+      _this.imagesData[dynamicDataObj.id].images = content.map(function(imgUrl) {
```
tonytlwu, line 2428:
```suggestion
      _this.imagesData[dynamicDataObj.id].images = _.map(content, function(imgUrl) {
```

js/utils.js
```diff
@@ -961,6 +965,16 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
           _.set(data, ['record', 'data', data.field.column], '');
         }
 
+        if (data.field.from === 'details') {
+          if (!Array.isArray(data.record.data[data.field.column])) {
+            data.record.data[data.field.column] = [];
```
tonytlwu, line 970: Given I'm reading this right, if the data is not an array, you're turning it into an empty array. What if there was a value but it's not an array. Do you end up wiping the value and replacing it with an empty array? Does this not have any repercussion in losing the data?

tonytlwu (changes requested):

js/layout-javascript/agenda-code.js
YaroslavOvdii: This logic was taken from the previously made [PR](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/425/files#diff-e51aca349bba35deb7224011da8726ab0fd50671ebeaba82184e04766f5b41beR1723) and now when I pasting images URL to the data source they most of the time separated with the new line.
Should we change that and if so what separator would be better to use?

js/utils.js
YaroslavOvdii: Yes, you are right it may lead us to lose the data.
But there is another issue.
Without this change, when users are setting the type to `image` and selecting the data type `image folder` the record in this `data.record.data[data.field.column]` object stays empty and we can't display images that in the folder.
Is there any way we could add images from the folder to this object `data.record.data[data.field.column]` without this change?

js/layout-javascript/agenda-code.js
tonytlwu: Thanks for the reference. Looks like it was the first commit of [this PR](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/425/commits) in Aug 2020. I read through the requirements and there's no such a requirement. Plus, if the value is a number e.g. media ID, then `1.split('\n')` will create an error.

I therefore don't think there's a need to split the value by linebreaks or anything.

js/utils.js
tonytlwu: I still don't understand why it's a problem for `data.record.data[data.field.column]` to stay exactly as it is, which could be a string, number, undefined or anything else.

I also see a few lines up that there's a section of code like this:

```js
        var image = _.get(data, ['record', 'data', data.field.column]);

        if (_.isArray(image)) {
          image = _.compact(image)[0];
        }

        if (isValidImageUrl(image)) {
          // Record data doesn't need updating
          return data.record;
        }
```

This suggests that there are scenarios where an array of images will be changed to only show the first one. That doesn't sound like this meets our requirements.

tonytlwu (changes requested):

js/layout-javascript/agenda-code.js
YaroslavOvdii: @tonytlwu When users add images through the form builder we already receiving an array of the image URLs and I didn't see cases when we receive a media ID.
And what about cases when users put image URLs manually in the data source?
In that case, we will receive a string with URLs

js/utils.js
YaroslavOvdii: @tonytlwu I've made some [screencasts](https://share.getcloudapp.com/o0uEQ895) with voice over where I'm trying to explain why I've done those changes and the code above. I hope it will make something clearer.

YaroslavOvdii:
@tonytlwu Added changes we have discussed earlier.

YaroslavOvdii:
@tonytlwu 
About this comment https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/520/#discussion_r569606032
We will not change this part at all in our new solution

templates/build/small-card-detail.build.hbs
```diff
@@ -54,12 +54,21 @@
         \{{#ifCond type '==' 'image'}}
           \{{#if content}}
             \{{#if labelEnabled}}
-            <div class="small-card-list-detail-label">\{{label}}</div>
+	            <div class="small-card-list-detail-label">\{{label}}</div>
```
tonytlwu, line 57:
```suggestion
              <div class="small-card-list-detail-label">\{{label}}</div>
```

js/utils.js
```diff
@@ -886,36 +890,48 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+              return true;
+            }
+          });
+
+          if (!urlEdited) {
+            _.set(data, ['record', 'data', data.field.column], '');
           }
-        });
+        } else if (data.field.from === 'details') {
+          var filterImageFiles = [];
```
tonytlwu, line 926:
```suggestion
          var imageFiles = [];
```

js/utils.js
```diff
@@ -886,36 +890,48 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
           return data.record;
         }
 
-        var image = _.get(data, ['record', 'data', data.field.column]);
+        if (data.field.from === 'summary') {
```
tonytlwu, line 893: The current implementation doesn't rely on any checks of `data.field.from`. You have now added checks against `"summary"` and `"detail"`. Is there any possibility where the `data.field.from` could be neither `"summary"` or `"details"`? An empty/null value perhaps?

Is it safer to just go with `if (data.field.from === 'detail') { /* ... */ } else { /* ... */ }`?

js/layout-javascript/small-h-card-code.js
```diff
@@ -338,6 +339,13 @@ DynamicList.prototype.attachObservers = function() {
       }).then(function() {
         Fliplet.UI.Actions(options);
       });
+    })
+    .on('click', '.multiple-images-item', function() {
+      var id = $(this).parent().data('detailEntryId');
+
+      _this.imagesData[id].options.index = $(this).index();
```
tonytlwu, line 346: https://github.com/airbnb/javascript#jquery--cache

js/layout-javascript/agenda-code.js
```diff
@@ -2403,7 +2411,37 @@ DynamicList.prototype.addDetailViewData = function(entry) {
       content = entry.originalData[dynamicDataObj.column];
     }
 
-    content = _this.Utils.String.toFormattedString(content);
```
tonytlwu, line 2406: What happened to what `_this.Utils.String.toFormattedString()` used to do? I assume we still need the ability to format `content` appropriately into string. I believe we introduced it to make sure arrays are correctly joined with `", "`.

js/layout-javascript/agenda-code.js
```diff
@@ -2403,7 +2411,37 @@ DynamicList.prototype.addDetailViewData = function(entry) {
...
+    if (dynamicDataObj.type === 'image') {
+      content = entry.originalData[dynamicDataObj.column];
+
+      var contentArray;
+
+      if (typeof content === 'string') {
+        var detectURLRegex = /((?:ftp|http|https):\/\/(?:\w+:{0,1}\w*@)?(?:\S+)(?::[0-9]+)?(?:\/|\/(?:[\w#!:.?+=&%@!-/]))?)/;
```
tonytlwu, line 2420: Please add a comment to explain what this RegExp is meant to do.

I also see this process repeated several times in all the layout JS files. Can we consolidate them to keep it DRY?

tonytlwu (changes requested):

js/utils.js
YaroslavOvdii: @tonytlwu As far as I tested I didn't see any other option except `summary` or `detail`.
Therefore I'll go with this `if (data.field.from === 'detail') { /* ... */ } else { /* ... */ }`.

js/layout-javascript/agenda-code.js
YaroslavOvdii: @tonytlwu I'm not sure why this line was removed. But I believe it shouldn't. I'll move it back.

YaroslavOvdii:
@tonytlwu added suggestions

js/utils.js
tonytlwu: > Therefore I'll go with this `if (data.field.from === 'detail') { /* ... */ } else { /* ... */ }`.

@YaroslavOvdii You say this, but you haven't committed any code to apply the changes.

YaroslavOvdii: @tonytlwu remade if statement.

templates/build/small-card-detail.build.hbs
```diff
@@ -54,12 +54,21 @@
         \{{#ifCond type '==' 'image'}}
           \{{#if content}}
             \{{#if labelEnabled}}
-            <div class="small-card-list-detail-label">\{{label}}</div>
+	          <div class="small-card-list-detail-label">\{{label}}</div>
```
tonytlwu, line 57: Incorrect indentation

js/layout-javascript/agenda-code.js
tonytlwu: @YaroslavOvdii I don't see this being responded to or dealt with. I still see this being repeated a lot and wonder if it can be modularised through `utils.js`.

tonytlwu (changes requested):

YaroslavOvdii:
@tonytlwu [fixed](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/520#discussion_r603292417)

css/interface.css
```diff
@@ -992,6 +992,10 @@ header .btn,
   width: 18%;
 }
 
+.detail-table-panels-holder .image-type-select .rTableCell.title {
+  width: 20.6%;
```
tonytlwu, line 996: @YaroslavOvdii Why 20.6% specifically?

js/layout-javascript/news-feed-code.js
```diff
@@ -2339,7 +2361,37 @@ DynamicList.prototype.addDetailViewData = function(entry) {
...
     }
 
-    content = _this.Utils.String.toFormattedString(content);
+    if (obj.type === 'image') {
+      content = entry.originalData[obj.column];
+
+      var contentArray;
+
+      if (typeof content === 'string') {
+        contentArray = [];
```
tonytlwu, line 2370: This line seems unnecessary.

```suggestion
```

templates/build/small-card-detail.build.hbs
tonytlwu: @YaroslavOvdii This is not addressed. I commented on it 25 days ago cc @sofiiakvasnevska

js/layout-javascript/agenda-code.js
```diff
@@ -2547,7 +2569,38 @@ DynamicList.prototype.addDetailViewData = function(entry) {
       content = entry.originalData[dynamicDataObj.column];
     }
 
-    content = _this.Utils.String.toFormattedString(content);
+    if (dynamicDataObj.type === 'image') {
```
tonytlwu, line 2572: Everything inside this `if` clause seems to be repeated across all layouts too. You only made the regex part DRY with the `getImagesUrlsByRegex` function but I feel that there's a lot more that can be modularised.

tonytlwu (changes requested):

css/interface.css
tonytlwu: @YaroslavOvdii That doesn't explain why it's `20.6%`. How was this calculated?

YaroslavOvdii:
@tonytlwu opened a new [PR ](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/562) with all previous.

css/interface.css
YaroslavOvdii: @tonytlwu remade this solution without `20.6%`

PR #521, opened 2021-01-28
OD-321 [Fix] Make sort dropdown not cut by parent accordion

ivandevupp:
@sofiiakvasnevska
## Issue
OD-321 https://weboo.atlassian.net/browse/OD-321
## Description
Make sort dropdown not cut by parent accordion
## Screenshots/screencasts
https://monosnap.com/file/Ttqofrzhf9ET0Re24dZBL8sVnpO5Kn
## Backward compatibility
## Reviewers 
@upplabs-alex-levchenko @YaroslavOvdii


js/layout-javascript/small-card-code.js
```diff
@@ -132,9 +132,13 @@ DynamicList.prototype.attachObservers = function() {
...
   });
 
-  $(window).resize(function() {
-    _this.Utils.DOM.adjustAddButtonPosition(_this);
-  });
+  $(window)
+    .resize(function() {
+      _this.Utils.DOM.adjustAddButtonPosition(_this);
+    })
+    .on('show.bs.dropdown', function() {
```
tonytlwu, line 139: What if the LFD detail view also has Bootstrap dropdowns implemented? It seems that we only want to implement this for the sort dropdown, not all dropdowns.

tonytlwu (changes requested):

js/layout-javascript/small-card-code.js
tonytlwu: Also, please add inline comments for clarification

tonytlwu:
@ivandevupp Do we not need this for other LFD layouts that support the sorting UI?

js/layout-javascript/small-card-code.js
ivandevupp: If LFD has also Bootstrap dropdown they will also become cut by accordion, that's why I think we should implement this not just for sort dropdowns.
Added inline comments for clarification.

ivandevupp: Also, I have solution for accordion heading widget with only styles 
image

tonytlwu: @ivandevupp Thanks. That makes sense. In this case your solution has a few problems.

The `show.bs.dropdown` and `hide.bs.dropdown` event listeners should be applied to `_this.$container` as the `attachObservers()` function already has. This is so that:

1. The behaviour is restricted to LFDs
2. The behaviour is restricted to the specific instance of LFD

When you apply the `overflow` style, you shouldn't use `$('[data-collapse-id]').find('div.in')` as this looks for all collapses in the entire page. Instead, look for the parent accordion that contains the LFD, e.g. something like  `_this.$container.parents('[data-collapse-id]').find('div.in')`.

ivandevupp: I don't know why, but at the first time I couldn't attach observer to container, it didn't fire, that's why I attached it to window, but now it works.
Fixed for SmallCard/SimpleList/NewsFeed.

js/layout-javascript/news-feed-code.js
```diff
@@ -144,6 +144,20 @@ DynamicList.prototype.attachObservers = function() {
   });
 
   _this.$container
+    .on('show.bs.dropdown', function(e) {
+      var parentAccordion = $(e.currentTarget).parents('[data-collapse-id]')[0];
```
tonytlwu, line 148: https://github.com/airbnb/javascript/blob/master/README.md#jquery--dollar-prefix

```suggestion
      $(e.currentTarget).parents('[data-collapse-id]').style('overflow', 'visible');
```

Why did you choose to use `e.currentTarget` instead of `e.target`?

tonytlwu (changes requested):

js/layout-javascript/news-feed-code.js
ivandevupp: Because `e.currentTarget` is child of needed DOM element, but now I understand that these makes no sense when using `.parents()` function.
Fixed and added fix to prevent overlapping the dropdown:
image

tonytlwu (changes requested):
Does not meet coding standards

PR #522, opened 2021-01-28
OD-340 [Fix] Split row data by commas and check intersection for Is one of

ivandevupp:
@sofiiakvasnevska
## Issue
OD-340 https://weboo.atlassian.net/browse/OD-340
## Description
Split row data by commas and check intersection for Is one of
## Screenshots/screencasts
https://monosnap.com/file/AZq0twTFae3QHxC2gBkTu19ErI4AiP
## Backward compatibility
## Reviewers 
@upplabs-alex-levchenko @AndrRyaz @YaroslavOvdii


js/utils.js
```diff
@@ -500,7 +500,7 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
         }
 
         if (condition === 'oneof') {
-          return splitByCommas(filter.value).includes(rowData);
+          return Boolean(_.intersectionWith(splitByCommas(filter.value), splitByCommas(rowData), _.isEqual).length);
```
tonytlwu, line 503:
```suggestion
          return !!_.intersectionWith(splitByCommas(filter.value), splitByCommas(rowData), _.isEqual).length;
```

tonytlwu (changes requested):

js/utils.js
ivandevupp: FIxed.

ivandevupp: Also, added fix for filter values in double quotes, they didn't work as expected because of comma splitting of the row data.

tonytlwu (changes requested):
This is not necessary. The behaviour outlined in https://weboo.atlassian.net/browse/OD-340 is expected and not a bug. The code you added below is in fact a different kind of logic entirely. It's what we would consider **Contains one of**, which is something we're planning to have in the backlog cc @sofiiakvasnevska 

```js
!!_.intersectionWith(splitByCommas(filter.value), splitByCommas(rowData), _.isEqual).length
```

See https://weboo.atlassian.net/browse/OD-340?focusedCommentId=15448 for more.

PR #523, opened 2021-01-29
OD-86 [Feature] LFD should detect and print arrays as a list

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-86 https://weboo.atlassian.net/browse/OD-86

## Description

- Added global function to print array as a list in preview mode.

## Screenshots/screencasts
http://somup.com/crVUQBom0n
## Backward compatibility
## Reviewers


js/layout-javascript/agenda-code.js
```diff
@@ -886,6 +886,12 @@ DynamicList.prototype.initialize = function() {
     .then(function(records) {
       _this.listItems = _this.getPermissions(records);
 
+      _this.listItems.map(function(item) {
```
tonytlwu, line 889: In the few lines above this, we have `_this.Utils.Records.prepareData()` that looks through all the records to prepare them.

Then, you've got, in this section, code that loops through all the records **again** and then looping through all the properties within them to reassign them. This is highly inefficient.

If you want to pre-process something, do it within the `_this.Utils.Records.prepareData()` function -- though I don't see why that would be necessary for this feature.

js/layout-javascript/agenda-code.js
```diff
@@ -1227,6 +1233,8 @@ DynamicList.prototype.addSummaryData = function(records) {
         content = _this.convertTime(content);
       }
 
+      content = _this.Utils.DataSource.formatCellValue(content);
```
tonytlwu, line 1236:
```suggestion
      content = _this.Utils.String.toFormattedString(content);
```

`_this.Utils.String.toFormattedString()` should:

1. If the input is a string, return the string
2. If the input is undefined or null, return an empty string
3. If the input is an array, remove the following values and then return the value after join them with `, `
3. If the input is an object, return the value using `JSON.stringify()`

tonytlwu (changes requested):
In the video 1:06, the data was a string entered into the data source manager that looks like an array. It appears that the LFD is then processing the string into an array. There's no requirement to parse a string that looks like an array into an array.

As the ticket suggests, we're only concerned with what happens if the data is already saved as an array, i.e. saved via checkboxes in the form builder.

js/layout-javascript/agenda-code.js
AndrRyaz: @tonytlwu, fixed

js/utils.js
```diff
@@ -1596,6 +1596,26 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
     });
   }
 
+  function toFormattedString(options) {
```
tonytlwu, line 1599: Adds JSDoc

js/utils.js
```diff
@@ -1596,6 +1596,26 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+    if (!options) {
```
tonytlwu, line 1600: `0` should be returned as `"0"` not `""`.

Add a string type check to the top so that if it's a string the code doesn't need to go through all the other logic

js/utils.js
```diff
@@ -1596,6 +1596,26 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+
+    if (Array.isArray(options)) {
+      options = options.filter(function(item) {
+        return !!item.trim();
+      });
+
+      options = options.join(', ');
+    }
+
+    if (typeof options === 'object') {
```
tonytlwu, line 1612: Use `else if`

js/utils.js
```diff
@@ -1596,6 +1596,26 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+      return JSON.stringify(options);
+    }
+
+    return options;
```
tonytlwu, line 1616: In this case, a number will be returned as number. A number should be returned as the number in string format.

tonytlwu (changes requested):

js/utils.js
```diff
@@ -1596,6 +1596,26 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
   }
 
+  function toFormattedString(options) {
+    if (!options) {
+      return '';
+    }
+
+    if (Array.isArray(options)) {
+      options = options.filter(function(item) {
+        return !!item.trim();
```
tonytlwu, line 1606: If there's an `undefined` in the array, this will trigger an error.

image

AndrRyaz: @tonytlwu, fixed

js/utils.js
```diff
@@ -1596,6 +1596,35 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+  function toFormattedString(options) {
+    switch (typeof options) {
+      case 'string':
+        return options;
+      case 'number':
+      case 'boolean':
+        return options.toString();
+      case 'object':
+        if (Array.isArray(options)) {
+          options = options.filter(function(item) {
```
tonytlwu, line 1614: I think a better way is to use

```js
options = _.filter(_.map(options, toFormattedString), function(part) { return part.trim().length });
```

This does two things:

1. Each value in the array could be of any type. This treats each value in the array with the same logic as the other values.
2. `undefined` and `null`, which results in `""` is removed instead of outputting `, ,`

tonytlwu (changes requested):

js/layout-javascript/agenda-code.js
tonytlwu: Point 2 specified:

> If the input is undefined or null, return an empty string

image

js/utils.js
AndrRyaz: @tonytlwu, fixed

AndrRyaz: @tonytlwu, fixed

js/layout-javascript/agenda-code.js
AndrRyaz: @tonytlwu, fixed
изображение

tonytlwu (approved):

Risk: High
Affected: LFD summary and detail view
Verify: All summary and detail views are correctly rendered

PR #532, opened 2021-02-08
OD-347 [Fix] Users data source will behaive as expexted

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-347

## Description
The bug occurred when users were trying to add comments and without saving, add entry management. 
Before this fix we didn't connect `this.config.social.comments` with checkbox `#enable-comments` on change event. Therefore when we tried add entry managment before we save settings `this.config.social.comments` still was `false` even if we have selected it earlier.
This fix removes that bug.

## Screenshots/screencasts
https://share.getcloudapp.com/GGu6klgz

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


squallstar (approved):

- **Risk**: Low
- **Affected**: Studio users configuring the component
- **Verify**: Verify the issue is resolved

PR #535, opened 2021-02-10
OD-357 [Fix] Users with previosly setup filters will see correct selected filter type and value

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-237

## Description
That seems to be when old users receive updates they don't have an `item.fieldValue` or `item.valueType` variables in the filter option.
That is why we were seeing empty `Value type` and `Value` fields. 
With the current fix, we will have a fullback in cases when users have an old configuration of the filter options.

## Screenshots/screencasts
Screenshot_23

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

## Notes
Unfortchunotly I wasn't able to replicate this issue on staging because I don't have, an old configurated LFD with my package to see changes but I'm sure this fix will work on staging.


tonytlwu:
@YaroslavOvdii 

> Unfortunately, I wasn't able to replicate this issue on staging because I don't have, an old configurated LFD with my package to see changes but I'm sure this fix will work on staging.

I suggest you find the point in the commit history before the last set of changes were applied so you can set up widgets in Studio that have the old configuration settings and then upload your fix to confirm it works. We can't have bug fixes that you haven't verified works.

PR #537, opened 2021-02-15
OD-357 [Fix] Users with previosly setup filters will see correct selected filter type and value

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-357 https://weboo.atlassian.net/browse/OD-357

## Description
That seems to be when old users receive updates they don't have an item.fieldValue or item.valueType variables in the filter option.
That is why we were seeing empty Value type and Value fields.
## Screenshots/screencasts
изображение

## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


tonytlwu (approved):

Risk: High
Affected: Studio users configuring filter options
Verify: Filter options are saved and loaded to work correctly

Squash and merge: Select "Enter a value" type by default

PR #538, opened 2021-02-16
OD-375 [Fix] User is unable to save filter configurations

AndrRyaz:
@sofiiakvasnevska 

## Issue
OD-375 https://weboo.atlassian.net/browse/OD-375

## Description
User is not able to save the filter configurations because there is a wrong if condition for checking the entry value from input
## Screenshots/screencasts
http://somup.com/crnDX3DYot
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


tonytlwu (approved):

Risk: Medium
Affected: Studio users configuring filter conditions
Verify: Filter options can be saved and loaded to work correctly

Squash and merge: Allow filters be saved with "none" values

js/interface-lists.js
```diff
@@ -506,7 +506,7 @@ function attahObservers() {
               return;
             }
 
-            if (!validate(field.value)) {
+            if (!field.value || !field.value.trim()) {
```
squallstar, line 509: what if `field.value` isn't a string?

AndrRyaz: @squallstar hi, this variable can only be a string with empty value or not empty

sofiiakvasnevska:
deployed to Production on 18/02/21

PR #543, opened 2021-02-19
ID-710 [Fix] Leave Handlebars.SafeString objects for templates to render

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-710

The last fix https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/533 attempted to prevent Handlebars.SafeStrings from being converted to string with `JSON.stringify()`. However, the bug fix was incorrect. It shouldn't have been forced into string with `Object.prototype.toString()`. Instead, it should have been kept as it is so that Handlebars templates can process it as a safe string.


squallstar (approved):

sofiiakvasnevska:
deployed to production 19/02/21

PR #558, opened 2021-03-31
OD-426 [Fix] Users will be able to see box-shadows after LFD layout selection without reloading the page.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-426

## Description
When users will select the LFD layout `data-settings-layout` attribute will be added right away.

## Screenshots/screencasts
https://share.getcloudapp.com/5zuB7yPJ

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko @ivandevupp @AndrRyaz


tonytlwu (changes requested):
The current approach means that `_this.Utils.DOM.addLayoutData()` is always executed but is rarely necessary. Each instance only ever needs it once.

I'd do this instead.

1. When the user selects a layout in the interface, send an event to the app preview ([link](https://developers.fliplet.com/API/core/studio.html#sending-data-from-widget-interface-to-app-preview))
2. `build.js` can include a handler so that when a layout is selected, the layout is added as `data-settings-layout`.

YaroslavOvdii:
@tonytlwu  I've encountered some issues when I've tried to implement your suggestion.
It seems that when I fired this function in the inteface.js
```
          Fliplet.Studio.emit('page-preview-send-event', {
            type: 'layout',
            data: {
              layout: listLayout
            }
          });
``` 
And trying to catch it with this function in the build.js
```
  Fliplet.Studio.onEvent(function(event) {
    if (event.type === 'layout') {
      var $mainContainer = $('[data-dynamic-lists-id="' + id + '"]').parent();

      if (!$mainContainer.attr('data-settings-layout')) {
        $mainContainer.attr('data-settings-layout', event.data.layout);
      }
    }
  });
```
There is nothing happens, because of my event.type never getting through.
Here is a [video](https://share.getcloudapp.com/bLuAJDL9) as well I hope it will help with clarification of what I'm doing.
Could you please guide me in this situation?

tonytlwu:
@YaroslavOvdii See below. You should have been able to debug this yourself.

image

YaroslavOvdii:
@tonytlwu I'm still not able to go thew.
Please see the [video](https://share.getcloudapp.com/5zuBD1Z0), that my event is still not calling.

tonytlwu:
@YaroslavOvdii My video here shows that the event sending & handling mechanism seems to be working fine. https://share.getcloudapp.com/eDujo12G

Your video shows the `Fliplet.Studio.onEvent()` being paused several times due to events being fired. But there's no sign of any breakpoints/debugger or pauses happening when the event is sent. Maybe it was never sent and therefore it isn't working.

Also, in `build.js` you're adding `Fliplet.Studio.onEvent()` inside the `Fliplet.Widget.instance('dynamic-lists', fn)` function. I wouldn't do that. If you do this, there will be as many event handlers as there are instances on the page. We only really need one handler.

YaroslavOvdii:
> Also, in build.js you're adding Fliplet.Studio.onEvent() inside the Fliplet.Widget.instance('dynamic-lists', fn) function. I wouldn't do that. If you do this, there will be as many event handlers as there are instances on the page. We only really need one handler.

@tonytlwu 
But to correctly determinate the LFD container on which we need additional data attribute we will need instance `this` at least in my solution. 
Because as far as I understand at the time when build.js is launched, the is no Utils.js or any other file where we could put function that can have access to the widget instance container is that right?

tonytlwu:
@YaroslavOvdii When you send the event you can send the instance ID too, then the handler can decide what to do based on that information.

```js
          Fliplet.Studio.emit('page-preview-send-event', {
            type: 'layout',
            data: {
              layout: listLayout,
              id: widgetInstanceId
            }
          });
```

YaroslavOvdii:
@tonytlwu Thanks, I've completely forgotten about that. And you are right all my issues were because I've declared listener inside the instance when I moved it out it worked as it should. In 10 minutes I'll add a new commit to this PR.

YaroslavOvdii:
@tonytlwu added a new solution using Studio events.
New [screencast](https://share.getcloudapp.com/WnuYwrO0)

js/interface.js
```diff
@@ -154,6 +154,14 @@ var DynamicLists = (function() {
           listLayout = $(this).data('layout');
           isLayoutSelected = true;
 
+          Fliplet.Studio.emit('page-preview-send-event', {
+            type: 'layout',
```
tonytlwu, line 158: Event types are shared across all apps and features so we should keep this specific to the LFD.

```suggestion
            type: 'dynamicListLayout',
```

build.html
```diff
@@ -77,7 +77,7 @@
     </script>
   {{/if}}
 {{else}}
-<div class="dynamic-list-placeholder">
+<div class="dynamic-list-placeholder" data-dynamic-lists-id="{{ id }}">
```
tonytlwu, line 80: There's no need to update this.

```suggestion
<div class="dynamic-list-placeholder">
```

js/build.js
```diff
@@ -8,3 +8,11 @@ Fliplet.Widget.instance('dynamic-lists', function(data) {
     dynamicLists[id] = new DynamicLists(data, container);
   }
 });
+
+Fliplet.Studio.onEvent(function(event) {
+  var eventDetail = event.detail;
+
+  if (eventDetail.type === 'layout') {
+    $('[data-dynamic-lists-id="' + eventDetail.data.id + '"]').parent().attr('data-settings-layout', eventDetail.data.layout);
```
tonytlwu, line 16: You can make use of the existing attributes.

```suggestion
    $('.fl-widget-instance[data-id="' + eventDetail.data.id + '"]').attr('data-settings-layout', eventDetail.data.layout);
```

tonytlwu (changes requested):

YaroslavOvdii:
@tonytlwu added suggestions

tonytlwu (commented):
@YaroslavOvdii Why have the interface and build templates changed? It's likely they are insignificant changes like `\r` etc. but I'd prefer not to include any because we can't be certain the changes are safe without carefully looking through all the changes.

YaroslavOvdii:
@tonytlwu I've rebuilt templates with `fliplet run` and `\r` was removed, but still build templates have changed and as I looked at them they are odd because it seems that line was removed and then exact the same line was added once more. Is that ok?

tonytlwu:
@YaroslavOvdii You can restart the work on a different branch and then make sure the template files are not committed.

YaroslavOvdii:
New [PR](https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/561) was opened.

PR #563, opened 2021-04-06
OD-470 [Fix] Show applied filters when applying them from the filters overlay

AndrRyaz:
@sofiiakvasnevska 

## Issue
OD-470 https://weboo.atlassian.net/browse/OD-470
## Description
Added condition to show active filter when applying them from overlay
## Screenshots/screencasts
http://somup.com/crf16qbdRp
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko @ivandevupp @YaroslavOvdii


js/layout-javascript/agenda-code.js
```diff
@@ -203,6 +203,10 @@ DynamicList.prototype.attachObservers = function() {
       $(this).parents('.new-agenda-search-filter-overlay').removeClass('display');
       _this.$container.find('.section-top-wrapper, .agenda-cards-wrapper, .dynamic-list-add-item').removeClass('hidden');
 
+      if (!_this.$container.find('.clear-filters').hasClass('hidden')) {
```
tonytlwu, line 206: What exactly is this `if` clause logic looking to confirm?

AndrRyaz: @tonytlwu I came up with such an output in order to see if we have selected filters, because at the moment the Appli button is pressed, there are no active filters in the variable `_this.activeFilters` and the same for filters in active filters container, so this is the way how i can fix this bug

tonytlwu: @AndrRyaz If you need to check if we have selected filters, then check if we have selected filters.

Something like `this.$container.find('.hidden-filter-controls-filter.mixitup-control-active')` should be able to check what's selected.

AndrRyaz: @tonytlwu fixed

tonytlwu (approved):

Risk: Medium
Affected: LFD layouts with filter overlay enabled
Verify: Filter overlays are workined correctly

PR #571, opened 2021-04-15
OD-382 [Feature] LFD update with date formats using moment locales

ivandevupp:
@sofiiakvasnevska
## Issue
OD-382 https://weboo.atlassian.net/browse/OD-382
## Description
LFD updated with date formats using moment locales for NewsFeed (Layout, Comments), Agenda (Layout) and Interface (data view settings)
## Screenshots/screencasts
https://monosnap.com/file/V6FWD6TBTjR6GTZqEiGk54ZtSn3sDM
## Backward compatibility
## Reviewers 
@upplabs-alex-levchenko @AndrRyaz @YaroslavOvdii


js/interface.js
```diff
@@ -63,6 +63,12 @@ var DynamicLists = (function() {
   var defaultColumns = window.flListLayoutTableColumnConfig;
   var defaultEntries = window.flListLayoutTableConfig;
 
+  var LOCALE_FORMATS = {
+    TIME: moment.localeData().longDateFormat('LT'),
+    DATE: moment.localeData().longDateFormat('ll'),
+    DATE_TIME: moment.localeData().longDateFormat('lll')
```
tonytlwu, line 69: Where did you use `LOCALE_FORMATS.DATE_TIME`?

js/layout-javascript/agenda-code.js
```diff
@@ -989,6 +989,7 @@ DynamicList.prototype.initialize = function() {
     Fliplet.App.Storage.remove('flDynamicListQuery:' + _this.data.layout);
   }
 
+  moment.locale(navigator.language);
```
tonytlwu, line 992: You removed a line of `moment.locale('en')` in `utils.js`. Why didn't you just replace that line with this line here, instead of having to repeat this in every layout JS file?

js/utils.js
```diff
@@ -236,7 +239,7 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
         context = undefined;
       }
 
-      return getMomentDate(context).format(block.hash.format || 'DD MMMM YYYY');
+      return getMomentDate(context).format(moment()._locale._longDateFormat.ll || moment()._locale._longDateFormat.LL);
```
tonytlwu, line 242: 1. `moment()._locale._longDateFormat.ll` should be the same as `moment.localeData().longDateFormat('ll')`. Use the function instead of accessing the `_longDateFormat` object directly.
2. Why are you using `ll` with the `LL` fallback? Why not just reference `LOCALE_FORMATS.DATE`?
3. `block.hash.format` needs to be kept because it gives users a way to customise the output format. If you remove `block.hash.format` then existing Handlebars helpers that use a custom format will break.

js/utils.js
```diff
@@ -2299,7 +2302,8 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
       appendUrlQuery: appendUrlQuery
     },
     Date: {
-      moment: getMomentDate
+      moment: getMomentDate,
+      LOCALE_FORMATS
```
tonytlwu, line 2306: I'd rather use a method, e.g. `this.Utils.Date.getLocaleFormat(format)` with `format` would be `date` or `time`. Functions are easier to upgrade should it be necessary.

widget.json
```diff
@@ -84,6 +84,7 @@
       "js/build-lists.js",
       "js/build.js",
       "css/build.css",
+      "vendor/moment-with-locales.js",
```
tonytlwu, line 87: This shouldn't be necessary. The `moment` dependency should come with all the locale information already.

tonytlwu (changes requested):
Rejected +1

PR #574, opened 2021-04-15
ID-875 [Fix] Add invalid filters from the query to the filter UI

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-875

### Changelog

1. When filters from the query are invalid (i.e. not found in the data entries):
    1. The filter queries are shown as filter options to honour the original query
    2. **No matches found** is displayed
2. When search query returns no results, **No matches found** is displayed
3. Performance improvements added to use `_.get(foo, ['data', 'bar'])` instead of `_.get(foo, 'data.bar')`
4. Fix an issue where selected filters aren't shown when filter overlays are enabled
5. Fix an issue where if multiple selected filters are shown from the filter overlay, the filters aren't removed specifically for a field at a time
6. Fix an issue where agenda filters were hidden when filter queries are used
7. ESLint cleanup


js/utils.js
```diff
@@ -1106,19 +1107,70 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+        _.forEach(value, function(value) {
+          if (typeof value === 'undefined') {
+            return;
+          }
+
+          flFilters.push({
+            type: field,
+            data: {
+              name: value,
+              class: ('' + value).toLowerCase().replace(/[!@#\$%\^\&*\)\(\ ]/g, '-')
```
squallstar, line 1159: What's this regexp for? Leave a comment to explain what it does and matches

squallstar (approved):

js/utils.js
tonytlwu: @squallstar This was for creating kebab case for class names. I've upgraded this to use `_.kebabCase()`.

tonytlwu:
@squallstar This is ready for review again.

squallstar (approved):

PR #598, opened 2021-08-04
OD-537 [Fix] Prevent background scrolling in Directory and Featured List layouts

skrupnyk:

OD-537 https://weboo.atlassian.net/browse/OD-537

## Description
Fixed background scrolling in LFD detail view Featured List

## Screenshots/screencasts
https://storyxpress.co/video/krm0x018st3r4v6eu

## Backward compatibility

## Reviewers

### DIFF
--- a/js/layout-javascript/small-card-code.js
+++ b/js/layout-javascript/small-card-code.js
@@ -2028,13 +2028,14 @@ DynamicList.prototype.showDetails = function(id, listData) {
 
         var template = Handlebars.compile(data.src || src);
         var wrapperTemplate = Handlebars.compile(wrapper);
+        var iframeBody = $('.fl-page-content-wrapper').parent();
 
         // This bit of code will only be useful if this component is added inside a Fliplet's Accordion component
         if (_this.$container.parents('.panel-group').not('.filter-overlay').length) {
           _this.$container.parents('.panel-group').not('.filter-overlay').addClass('remove-transform');
         }
 
-        $('.fl-page-content-wrapper').parent().css('overflow', 'hidden');
+        iframeBody.css('overflow', 'hidden');
 
         // Adds content to overlay
         $overlay.find('.small-card-detail-overlay-content-holder').html(wrapperTemplate(entryId));
@@ -2074,11 +2075,12 @@ DynamicList.prototype.closeDetails = function() {
   // Function that closes the overlay
   var _this = this;
   var $overlay = $('#small-card-detail-overlay-' + _this.data.id);
+  var iframeBody = $('.fl-page-content-wrapper').parent();
 
   Fliplet.Page.Context.remove('dynamicListOpenId');
   $overlay.removeClass('open');
   _this.$container.find('.new-small-card-list-container').removeClass('overlay-open');
-  $('.fl-page-content-wrapper').parent().css('overflow', 'auto');
+  iframeBody.css('overflow', 'auto');
 
   setTimeout(function() {
     $overlay.removeClass('ready');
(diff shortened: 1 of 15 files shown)


js/layout-javascript/small-card-code.js
```diff
@@ -2027,12 +2028,15 @@ DynamicList.prototype.showDetails = function(id, listData) {
 
         var template = Handlebars.compile(data.src || src);
         var wrapperTemplate = Handlebars.compile(wrapper);
+        var $pageContentBody = $('.fl-page-content-wrapper').parent();
```
tonytlwu, line 2031: Why do we use `$('.fl-page-content-wrapper').parent()` and not `$('body')`?

tonytlwu (commented):
Why does this problem not affect the simple list or news feed layouts?

js/layout-javascript/small-card-code.js
skrupnyk: > Why do we use `$('.fl-page-content-wrapper').parent()` and not `$('body')`?

I use ```$('.fl-page-content-wrapper').parent()``` because ```$('body')``` is selected only for a specific list on the page, but not for all content in the iframe

tonytlwu: @skrupnyk I don't understand the scenario you are describing.

image

This is a snapshot of our engine template. As you can see, the container with `.fl-page-content-wrapper` is always the direct child of `<body>` if the app supports the latest D&D. If it doesn't, then there is no `.fl-page-content-wrapper`.

Using `$('.fl-page-content-wrapper').parent()` would either:

1. Always lead to `<body>` if the app supports the latest D&D; or
2. Lead to nothing if it doesn't support the latest D&D.

So why wouldn't we just reference the `body` element directly? What's the scenario you have in mind that means we should use `$('.fl-page-content-wrapper').parent()`?

skrupnyk: > @skrupnyk I don't understand the scenario you are describing.
> 
> image
> 
> This is a snapshot of our engine template. As you can see, the container with `.fl-page-content-wrapper` is always the direct child of `<body>` if the app supports the latest D&D. If it doesn't, then there is no `.fl-page-content-wrapper`.
> 
> Using `$('.fl-page-content-wrapper').parent()` would either:
> 
> 1. Always lead to `<body>` if the app supports the latest D&D; or
> 2. Lead to nothing if it doesn't support the latest D&D.
> 
> So why wouldn't we just reference the `body` element directly? What's the scenario you have in mind that means we should use `$('.fl-page-content-wrapper').parent()`?

understood, make sense

skrupnyk:
> Why does this problem not affect the simple list or news feed layouts?

no such problems were found in the simple list or news feed layouts

tonytlwu:
@skrupnyk I don't think that's a detailed way to answer the question. Why was no problem found in the simple list or news feed layouts? What's different about those layouts that meant it didn't affect them?

skrupnyk:
> @skrupnyk I don't think that's a detailed way to answer the question. Why was no problem found in the simple list or news feed layouts? What's different about those layouts that meant it didn't affect them?

the simple list or news feed layouts do not have such a problem, because they already have a ready-made solution and I am ready to submit my code as in these components. The solution for the simple list or news feed layouts seems to me the best.
Here is the solution - ```$('body').addClass('lock');```

tonytlwu:
@skrupnyk Let's use the same solution that the simple list and news feed layout have then. I don't see why we need to reinvent the wheel here.

Rejected +1

skrupnyk:
> @skrupnyk Let's use the same solution that the simple list and news feed layout have then. I don't see why we need to reinvent the wheel here.
> 
> Rejected +1

fixed

tonytlwu (approved):
@sofiiakvasnevska @inna-bieshulia This is ready for testing

Risk: Medium
Affected: Directory and Featured list layouts
Verify: The LFD list and detail views are appropriately scrollable on all devices

PR #599, opened 2021-08-10
OD-550 [Feature] Apply date filter as inline filter

S-Volodymyr:

OD-550 https://weboo.atlassian.net/browse/OD-550
Description

Apply date filter as an inline filter
Screenshots/screencasts

Backward compatibility
If the user will have a filter by date on the main view and a custom field date on the detail view - the date range filter will be displayed as the date filter

Reviewers

Affected tasks in this PR:
https://weboo.atlassian.net/browse/OD-550 (Apply date filter as inline filter)
https://weboo.atlassian.net/browse/OD-551 (Apply date filter as overlay filter)
https://weboo.atlassian.net/browse/OD-552 (Apply date filter using query parameters)
https://weboo.atlassian.net/browse/OD-553 (Apply appearance settings to date filters)
https://weboo.atlassian.net/browse/OD-557 (Apply event tracking to date filters)

TODO:
https://weboo.atlassian.net/browse/OD-554 (Apply localization to date filters)
https://weboo.atlassian.net/browse/OD-552 (Apply date filter using query parameters) - fix query
https://weboo.atlassian.net/browse/OD-553 (Apply appearance settings to date filters) - check settings
https://weboo.atlassian.net/browse/OD-557 (Apply event tracking to date filters) - change events names


js/layout-javascript/news-feed-code.js
```diff
@@ -284,6 +285,41 @@ DynamicList.prototype.attachObservers = function() {
         }, 0);
       }
     })
+    .on('click change keydown', '.hidden-filter-controls-filter-date-range', function(event) {
```
tonytlwu, line 288: Most of the other event listeners only listen for the `click` and `keydown` events. Why is this listening to the `change` event too?

templates/build/news-feed-filters.build.hbs
```diff
@@ -14,8 +14,27 @@
         <div id="collapse-\{{id}}-\{{@index}}" class="panel-collapse collapse">
           <div class="panel-body">
             <fieldset data-filter-group>
-              \{{#each [data]}}
-              <div class="btn hidden-filter-controls-filter focus-outline" tabindex="0" data-toggle="\{{[class]}}" data-field="\{{../name}}" data-value="\{{[name]}}" >\{{[name]}}</div>
+              \{{#each  ../detailViewOptions}}
+                \{{#ifCond type '==' 'date'}}
+                  \{{#each ../data}}
+                    \{{#ifCond ../../name '==' 'Date'}}
```
tonytlwu, line 20: Why are we checking for the "Date" name? I think the only requirement is for the type to be a date.

templates/build/news-feed-filters.build.hbs
```diff
@@ -14,8 +14,27 @@
         <div id="collapse-\{{id}}-\{{@index}}" class="panel-collapse collapse">
           <div class="panel-body">
             <fieldset data-filter-group>
-              \{{#each [data]}}
-              <div class="btn hidden-filter-controls-filter focus-outline" tabindex="0" data-toggle="\{{[class]}}" data-field="\{{../name}}" data-value="\{{[name]}}" >\{{[name]}}</div>
+              \{{#each  ../detailViewOptions}}
```
tonytlwu, line 17: Instead of using Handlebars to go through this logic (for which I feel makes it quite difficult to debug/update and enhance, I think the data going into the Handlebars template should be pre-processed so that Handlebars knows, for each filter, which type of filter it should render, e.g.

```js
var filters = [{ type: 'default', name: 'location' }, { type: 'date', name: 'DoB' }];
var tpl = Handlebars.compile('{{#each data}} {{#ifCond type '==' "date"}} <!-- Date filter markup --> {{else}} <!-- Standard filter markup --> {{/ifCond}} {{/each}}');
var html = tpl({ data: filters });
```

tonytlwu (changes requested):
1. I only see the news feed CSS file updated. The work should apply to all the layouts that support filters
2. I see agenda JS code updated, but I don't see any other files, e.g. templates or CSS, updated. This work seems incomplete
3. No screencast available for demo

See other inline comments.

Rejected +1

templates/build/news-feed-filters.build.hbs
S-Volodymyr: We checking for type from an additional field which is in the detail view section, and then we check for the Date filter from the filters setup. It was described here:
image

js/layout-javascript/news-feed-code.js
S-Volodymyr: we need _change_ event for the next situation:
**If the FROM date is updated to be later than to TO date (or the TO date is updated to be earlier than the FROM date), the TO (FROM) date is automatically updated to match the value of the FROM (TO) date**

templates/build/news-feed-filters.build.hbs
S-Volodymyr: Please advise, where it is possible to add type to the data set? Now we have only the name and class. And that's why I'm using the name of the column instead of type

templates/build/news-feed-filters.build.hbs
tonytlwu: Based on [this markup](https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/0179acd511a85a05c6f19c3ead98a64674397b67/templates/interface/detail-view-panels.interface.hbs#L27-L36) and [this code](https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/0179acd511a85a05c6f19c3ead98a64674397b67/js/layout-javascript/news-feed-code.js#L2381-L2387), it suggests that the detail view options are saved with a `type` property.

Changing the **Field type** dropdown should not lead to any change in any `name` property. If you find that is the case, please direct us to the code where a `name` property is being changed by updating the dropdown selection.

templates/build/news-feed-filters.build.hbs
tonytlwu: [This is the code](https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/0179acd511a85a05c6f19c3ead98a64674397b67/js/layout-javascript/news-feed-code.js#L1725-L1730) where the `filters` array (used for this template) is generated. I suggest passing the detail view options as a parameter to the `_this.Utils.Records.parseFilters()` function so that it can output the `filters` collection with a `type` property that's based on an assessment of the detail view options.

templates/build/news-feed-filters.build.hbs
```diff
@@ -14,8 +14,27 @@
...
+                      \{{#if @first}}
+                        <div class="date-picker-container">
+                          <i class="fa fa-calendar">&nbsp;</i>
+                          <input type="text" id="dateFrom" value="" class="date-picker btn overlay-filter hidden-filter-controls-filter-date-range focus-outline" data-toggle="dateFrom" data-field="\{{../../name}}" data-value="">
+                          to
+                        </div>
+                        <div class="date-picker-container">
+                          <i class="fa fa-calendar">&nbsp;</i>
+                          <input type="text" id="dateTo" value="" class="date-picker btn overlay-filter hidden-filter-controls-filter-date-range focus-outline" data-toggle="dateTo" data-field="\{{../../name}}" data-value="">
+                          <input type="button" value="Reset" id="resetDate" class="btn btn-link btn-link-reset hidden">
```
tonytlwu, line 30: 2 errors here.

1. You must not use unique `id`s in the HTML markup unless it is specific to the widget instance. That's because if we have multiple instances (or in this case, multiple date filters), then there will be multiple elements with `id="resetDate"` and that's invalid HTML. You can't have more that one element on the page with the same `id` as it would produce unexpected results.
2. Use kebab-case for HTML classes and IDs.

js/layout-javascript/news-feed-code.js
```diff
@@ -284,6 +285,41 @@ DynamicList.prototype.attachObservers = function() {
...
+
+      if ($filter.parents('.inline-filter-holder').length) {
+        // @HACK Skip an execution loop to allow custom handlers to update the filters
+        setTimeout(function() {
+          _this.searchData();
+        }, 0);
+      }
+    })
+    .on('click', '#resetDate', function() {
+      $('input.date-picker').datepicker('setDate', new Date());
```
tonytlwu, line 319: [Requirement](https://weboo.atlassian.net/browse/OD-550) says:

```
Scenario 2.6: Resetting filter by date range values

GIVEN the user selects the date "FROM"  and the date "TO" values

WHEN the user clicks the "Reset" link 

THEN the user sees the "FROM" and "TO" date pickers default to the minimum and maximum values found in the data loaded into the LFD 

AND the "Reset" link hides itself
```

This looks like the reset button is resetting the date pickers to the current date instead of to the minimum and maximum values found in the data loaded into the LFD.

js/layout-javascript/news-feed-code.js
tonytlwu: Where can I find the code that changes the TO date to match the FROM date (or vice versa)?

tonytlwu (changes requested):

PR #602, opened 2021-08-18
OD-550 [Feature] Apply date filter as inline filter

S-Volodymyr:

OD-550 https://weboo.atlassian.net/browse/OD-550
Description

Apply date filter as an inline filter
Screenshots/screencasts

Backward compatibility
If the user will have a filter by date on the main view and a custom field date on the detail view - the date range filter will be displayed as the date filter

Reviewers

Affected tasks in this PR:
https://weboo.atlassian.net/browse/OD-550 (Apply date filter as inline filter)

TODO:
Implement next points from the task description:
5. The FROM date must be the same or earlier than the TO date
>> compare selected dates. this part is ready, but should be refactored (1h)
6. If the FROM date is updated to be later than to TO date (or the TO date is updated to be earlier than the FROM date), the TO (FROM) date is automatically updated to match the value of the FROM (TO) date
>> event for datepickers was already added, we need just a comparison check (1h)
10. The date filter UI must be responsive to the available width
>> add condition into view (hbs) to display propper clas (it's already included into CSS) (~1h)


js/layout-javascript/news-feed-code.js
```diff
@@ -1739,16 +1777,44 @@ DynamicList.prototype.addFilters = function(records) {
...
     var template = _this.data.advancedSettings && _this.data.advancedSettings.filterHTML
       ? Handlebars.compile(_this.data.advancedSettings.filterHTML)
       : Handlebars.compile(filtersTemplate());
 
     _.remove(filters, function(filter) {
       return _.isEmpty(filter.data);
     });
+
     _this.$container.find('.filter-holder').html(template(filtersData));
+    _this.datePickerFrom = _this.$container.find('#dateFrom');
```
tonytlwu, line 1793: jQuery objects should use a `$` prefix.

```suggestion
    _this.$datePickerFrom = _this.$container.find('#dateFrom');
```

js/layout-javascript/news-feed-code.js
```diff
@@ -1739,16 +1777,44 @@ DynamicList.prototype.addFilters = function(records) {
...
+    _this.datePickerTo = _this.$container.find('#dateTo');
+
+    _this.datePicker = _this.$container.find('input.date-picker').datepicker({
```
tonytlwu, line 1796: I don't understand why this needs to be saved into a property.

js/layout-javascript/news-feed-code.js
```diff
@@ -1739,16 +1777,44 @@ DynamicList.prototype.addFilters = function(records) {
...
+      format: 'yyyy-mm-dd', // specific date format using lower case for Bootstrap datepicker
+      todayHighlight: true,
+      autoclose: true
+    }).on('changeDate', function(e) {
+      _this.value = moment(e.date).format(_this.DATE_FORMAT);
+    });
+
+    _this.datePickerFrom.datepicker({
```
tonytlwu, line 1804: This looks like the datepicker is being initiated twice.

js/layout-javascript/news-feed-code.js
```diff
@@ -1739,16 +1777,44 @@ DynamicList.prototype.addFilters = function(records) {
...
+      format: 'yyyy-mm-dd',
+      autoclose: true
+    }).on('changeDate', function(e) {
+      _this.datePickerTo.datepicker('setStartDate', new Date(e.date));
```
tonytlwu, line 1808: https://bootstrap-datepicker.readthedocs.io/en/latest/methods.html#setstartdate

This looks like it's setting a limit to the the date picker. We specifically said multiple times that users should not be limited on which dates they can pick.

js/utils.js
```diff
@@ -1300,12 +1301,20 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
           field: field,
           useData: true
         }), function(item) {
+          var dates = records.map(function(obj) {
+            return moment(obj.data.Date);
+          });
+          var minDate = moment.min(dates).format(LOCAL_FORMAT);
```
tonytlwu, line 1307: When there are invalid values, `moment.min()` will return the invalid value as the result. The same goes for `moment.max()`.

The invalid dates need to be excluded before calling `moment.min()` to avoid this problem.

image

js/utils.js
```diff
@@ -1300,12 +1301,20 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
           field: field,
           useData: true
         }), function(item) {
+          var dates = records.map(function(obj) {
```
tonytlwu, line 1304: This is already happening inside a `records.forEach()` loop, which means we're running a loop inside a loop. If the data source has 100,000 entries, which regularly happens, this would result in a lot of unnecessary processing time.

js/utils.js
```diff
@@ -1300,12 +1301,20 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
...
+          var maxDate = moment.max(dates).format(LOCAL_FORMAT);
           var classConverted = _.kebabCase(item);
           var newObj = {
             type: field,
             data: {
               name: item,
-              class: classConverted
+              class: classConverted,
+              type: field,
+              minDate: minDate,
```
tonytlwu, line 1316: I don't understand why every single record in the data set needs a `minDate` and `maxDate` property. Each date filter probably needs these properties, but not each record.

tonytlwu (changes requested):
Rejected +1

js/layout-javascript/news-feed-code.js
S-Volodymyr: Done. Thanks

PR #605, opened 2021-08-25
OD-566 [Fix] Required "User data" field if it is empty

skrupnyk:

OD-566 https://weboo.atlassian.net/browse/OD-566

## Description
Fixed "User data" field is required
**Problem:** Control which supposed to reflect validation lays inside of the iframe and does not fall into current selection.
**Solution:** Added specific case for data-source provider and selector updated to find it inside of the iframe

## Screenshots/screencasts
https://storyxpress.co/video/ksrkt7jinxb360kmx

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


skrupnyk:
**Problem: Control which supposed to reflect validation lays inside of the iframe and does not fall into current selection.
Solution: Added specific case for data-source provider and selector updated to find it inside of the iframe**

tonytlwu (changes requested):
Instead of a parent frame directly manipulating the markup/styles of the content inside an iframe, we prefer to send events/messages to the frame, and let the frame take care of it. This way we create a standard for how to send messages to the framed providers, e.g. send a validation error to the data source provider.

skrupnyk:
new PR https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/609

PR #607, opened 2021-09-03
OD-561 [FIX] Fixes an issue where "No security rules" warning appears multiple times

OleksandrLavrynenko:

OD-561https://weboo.atlassian.net/browse/OD-561

## Description
**Problem**: When the LDF component was opened, the "list-control" inputs were triggered. This emit an event to Data Source Provider and caused an alert
**Solution**: Removed the trigger for inputs. Added inputs setting activation if the input is active. Added accessRules generation to component initialization

## Screenshots/screencasts
https://storyxpress.co/video/kt49lq71577ye8jt7

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/interface.js
```diff
@@ -994,9 +999,25 @@ var DynamicLists = (function() {
...
 
           // Load Add. Edit, Delete
-          $('#add_entry').prop('checked', _this.config.addEntry).trigger('change');
-          $('#edit_entry').prop('checked', _this.config.editEntry).trigger('change');
-          $('#delete_entry').prop('checked', _this.config.deleteEntry).trigger('change');
+          $('#add_entry')
+            .prop('checked', _this.config.addEntry)
+            .parent()
+            .find('.hidden-settings')
+            .toggleClass('active', !!_this.config.addEntry);
```
tonytlwu, line 1006: It doesn't have to be a boolean.

```suggestion
            .toggleClass('active', _this.config.addEntry);
```

tonytlwu (changes requested):
I understand why `.trigger('change')` was removed. Why are the other changes necessary?

js/interface.js
OleksandrLavrynenko: > It doesn't have to be a boolean.

In our case _this.config.addEntry can be undefined and .toggleClass does not have any type conversions inside that's why we must convert it to boolean.

tonytlwu: @OleksandrLavrynenko Yes you're right. I just made the same mistake in another code that I was working on. We do need to convert it to boolean 👍🏻  Good work!

tonytlwu (approved):
@sofiiakvasnevska @inna-bieshulia This is ready for testing

Risk: Medium
Affected: Studio users configuring LFD
Verify: LFD data sources and security rules can be configured as expected

PR #611, opened 2021-09-15
ID-1373 [Feat] Add support for date range filter queries

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1373

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -203,8 +203,9 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
 
   /**
    * Determine filter types based on configuration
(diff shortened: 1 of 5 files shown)


squallstar (approved):

js/utils.js
```diff
@@ -2086,6 +2213,8 @@ Fliplet.Registry.set('dynamicListUtils', (function() {
   }
 
   function prepareRecordsData(options) {
+    debugger;
```
squallstar, line 2216: debugger to remove

squallstar (changes requested):

PR #642, opened 2021-11-12
OD-649 [Fix] Fixed default date assignment for LFD lists

OleksandrLavrynenko:
@romanyosyfiv

OD-649 https://weboo.atlassian.net/browse/OD-649

## Description
**Problem**: When using Hindi language, the default format "en" for moment.js was not used
**Solution**:  Added default format for moment.js

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/default-table-data/layouts-table-config.js
```diff
@@ -1,3 +1,5 @@
+moment.locale('en');
```
tonytlwu, line 1: I wouldn't fix it this way. If the moment locale is globally changed to another language e.g. `ar` then you've essentially undone the setting.

Aside from changing the global moment locale, moment objects can be individually changed to have different locales, e.g. `moment().locale('ar')` and `moment().locale('it')` are 2 moment objects with different locales (without changing the global locale setting.

You should find the default data used in this file (I think it applies to Agenda and News feed, but please verify all types) and use something like this: `moment().add(1, 'month').locale('en').format('YYYY-MM-DD')`.

tonytlwu, line 1: I wouldn't fix it this way. If the moment locale is globally changed to another language e.g. `ar` then you've essentially undone the setting.

Aside from changing the global moment locale, moment objects can be individually changed to have different locales, e.g. `moment().locale('ar')` and `moment().locale('it')` are 2 moment objects with different locales (without changing the global locale setting.

You should find the default data used in this file (I think it applies to Agenda and News feed, but please verify all types) and use something like this: `moment().add(1, 'month').locale('en').format('YYYY-MM-DD')`.

tonytlwu (changes requested):

js/default-table-data/layouts-table-config.js
OleksandrLavrynenko: **@tonytlwu**  Yes, I did it first. But I thought it didn't matter because we did the translations for the user interface.

tonytlwu (approved):

tonytlwu:
@romanyosyfiv @inna-bieshulia This is ready for testing

Risk: Low
Affected: Data sources created by LFD
Verify: Data is correctly created based on 

PR #654, opened 2021-12-06
OD-614 [Fix] Prevent list view from moving when opening & closing detailed views

OleksandrLavrynenko:
@romanyosyfiv

OD-614 https://weboo.atlassian.net/browse/OD-614

## Description
**Problem**: When the detailed view was opened, the LFD list was hidden and the content above the list remained on the page.
**Solution**: The LFD list is not hidden after opening the detailed view to display the page correctly and not to lose the scroll position. When the detailed view is opened, the close button is automatically focused. Focus blocked to items not included in the detailed view if it is open

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/layout-javascript/news-feed-code.js
```diff
@@ -379,10 +387,18 @@ DynamicList.prototype.attachObservers = function() {
...
         // find the element to expand and expand it
         if (_this.allowClick) {
-          $el.parents('.new-news-feed-list-container').addClass('hidden');
           _this.$container.find('.dynamic-list-add-item').addClass('hidden');
 
-          _this.showDetails(entryId);
+          _this.showDetails(entryId).then(function() {
+            setTimeout(function() {
+              _this.$closeButton.focus();
+              _this.$detailsContent.focusin(function() {
```
tonytlwu, line 395: Can you explain why the `.focusin()` and `.focusout()` event listeners are necessary?

js/layout-javascript/news-feed-code.js
```diff
@@ -2437,6 +2453,13 @@ DynamicList.prototype.showDetails = function(id, listData) {
   var wrapper = '<div class="news-feed-detail-wrapper" data-entry-id="{{id}}"></div>';
   var src = _this.src;
 
+  if (!this.$detailsContent || !this.$closeButton) {
+    this.$detailsContent = $('.news-feed-detail-overlay');
+    this.$closeButton = $(_.find(this.$detailsContent.find('.news-feed-detail-overlay-close'), function(element) {
```
tonytlwu, line 2458: See https://api.jquery.com/filter/

```suggestion
    this.$closeButton = this.$detailsContent.find('.news-feed-detail-overlay-close').filter(function(i, el) {
      return !$(el).hasClass('tablet');
    });
```

tonytlwu (changes requested):

js/layout-javascript/news-feed-code.js
OleksandrLavrynenko: When the focus disappears from the item in the popup, we shift the focus to the close buttonю.  But if the next focus element is the popup element, we undo the transition of the focus to the close button.
 
focusout =>  _this.$closeButton.focus();
focusin => cancellation of operation focusout if the next focus element is the popup element

tonytlwu: > When the focus disappears from the item in the popup, we shift the focus to the close button.

Very good idea. It wasn't discussed in our scenario planning, but this sounds like the right thing to do. It would be good to have this discussed and clarified first on Jira so QA also knows about it first. cc @romanyosyfiv @inna-bieshulia 

`_this.focusCloseButton(false)` doesn't make sense though. I took a look at the `focusCloseButton` function and if the parameter is `false`, the function simply does nothing. It doesn't look like any event listener is needed for `focusin`.

OleksandrLavrynenko: @tonytlwu _this.focusCloseButton(false) cancels focus on the close button when the next focus element is in the popup. For this I use debounce.

tonytlwu: @OleksandrLavrynenko This is where the `focusCloseButton()` is defined.

```js
DynamicList.prototype.focusCloseButton = _.debounce(function(isAction) {
  if (isAction) {
    this.$closeButton.focus();
  }
}, 200);
```

The debounced function is:

```js
function(isAction) {
  if (isAction) {
    this.$closeButton.focus();
  }
}
```

`_this.focusCloseButton(false)` calls the debounced function with `false` as its parameter.

```js
```js
function(isAction) {
  if (isAction) {
    this.$closeButton.focus();
  }
}
```

...does nothing if `isAction` is `false`.

Have I misunderstood something? It seems that this line does nothing.

OleksandrLavrynenko: **@tonytlwu** Yes, it does nothing but undo the previous action when true is passed. When `focusout `is triggered, the `_this.focusCloseButton (true)` function should shift the focus to the close button, but when `focusin `is triggered, the `_this.focusCloseButton (false)` function cancels this action. Thus, when the focus is on the last element and we press the Tab key, `focusout ` is started, and since the next element does not exist, this action is not canceled, and the focus goes to the close button.

tonytlwu: @OleksandrLavrynenko When `focusin` is triggered, how does `_this.focusCloseButton(false)` cancel the previous focus shift if `_this.focusCloseButton(false)` programmatically does nothing?

It looks like you can achieve the same result by just taking out the `focusin` event listener.

OleksandrLavrynenko: **@tonytlwu** No. If the `focusin `is off then the `focusout `will always assign focus to the close button

tonytlwu: @OleksandrLavrynenko I see what you tried to do after trying out the event listeners myself. It looks like `focusin` and `focusout` is occurring on all focusable elements inside the container, and not just the container. This means both `focusout` and `focusin` (in that specific order) are always being triggered.

Based on [this article](https://stackoverflow.com/questions/15418358/how-to-detect-when-a-container-loses-focus), I created the following solution which is cleaner and clearer at describing what's happening.

https://jsfiddle.net/tonytlwu/b3so9pLq/

Let me know if you disagree.

OleksandrLavrynenko: @tonytlwu That's cool! Thanks! I forgot about relatedTarget

js/layout-javascript/news-feed-code.js
```diff
@@ -392,12 +386,13 @@ DynamicList.prototype.attachObservers = function() {
...
-              }).focusout(function() {
-                _this.focusCloseButton(true);
+              _this.$detailsContent.focusout(function(event) {
+                if (event.currentTarget.contains(event.relatedTarget)) {
+                  return;
+                }
+    
+                _this.$closeButton.focus();
               });
-            }, 200);
```
tonytlwu: @OleksandrLavrynenko Why would you remove this line?

tonytlwu: @OleksandrLavrynenko The removal of this line means code wouldn't have worked since it becomes syntactically incomplete. I see the code has gone past interview review and marked as ready for testing. Please test your own work before submitting for review or testing. cc @upplabs-alex-levchenko

tonytlwu (changes requested):
Branch is out of date

js/layout-javascript/news-feed-code.js
```diff
@@ -58,6 +58,8 @@ function DynamicList(id, data) {
   this.sortOrder = 'none';
   this.sortField = null;
   this.imagesData = {};
+  this.$closeButton;
```
tonytlwu, line 61: This doesn't do anything.

Before this line, `this.$closeButton` is `undefined`. After this line, `this.$closeButton` is still `undefined`.

This means either you don't need these lines at all, or you should set them to `null`. I suggest simply remove them, unless there's a reason for these that I don't know about.

I know there are plenty of lines like this above as well and it looks like they're also unnecessary. You don't need to remove those now though. We don't want to introduce unnecessary risk.

tonytlwu (changes requested):

js/layout-javascript/news-feed-code.js
OleksandrLavrynenko: @tonytlwu I understand you. I added a default value of "null"

PR #657, opened 2021-12-23
ID-1462 [Feat] Number filters

tonytlwu:
Ref.

- https://weboo.atlassian.net/browse/ID-1462 Inline filter
- https://weboo.atlassian.net/browse/ID-1463 Overlay filter
- https://weboo.atlassian.net/browse/ID-1464 Query parameters
- https://weboo.atlassian.net/browse/ID-1466 Accessibility
- https://weboo.atlassian.net/browse/ID-1465 Event tracking

### Follow-up PRs

- ID-1778 [Fix] Clear number filters when all filters are cleared #659
- ID-1789 [Fix] Only exclude null, undefined or false values from filters #660
- ID-1796 [Fix] Ensures filters are correctly reset after using multiple filters #661
- ID-1804 [Fix] Avoid splitting filter values by comma if it's a date/number filter #664

PR #667, opened 2022-01-17
ID-1817 [Fix] Allow query filter to use "0"

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1817

Follows https://github.com/Fliplet/fliplet-widget-dynamic-lists/pull/657


tonytlwu:
Replaced fix with https://github.com/Fliplet/fliplet-api/pull/5114

PR #677, opened 2022-02-28
OD-749 [Fix] Fixed automatic scrolling LFD detail view

skrupnyk:
@romanyosyfiv @inna-bieshulia

OD-749 https://weboo.atlassian.net/browse/OD-749

## Description
**Problem:** The function of focusing on the close button is triggered when clicking anywhere in the popup in the detailed view.
**Solution:** Added a mousedown handler that calls the focus function if the action on the detail view popup is gone.

## Screenshots/screencasts
www.awesomescreenshot.com/video/7631676?key=26bef7b464e49742980a0e550b1c1af8

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/layout-javascript/small-card-code.js
```diff
@@ -2065,13 +2065,30 @@ DynamicList.prototype.showDetails = function(id, listData) {
         var template = Handlebars.compile(data.src || src);
         var wrapperTemplate = Handlebars.compile(wrapper);
 
+        function focusCloseBtn() {
```
tonytlwu, line 2068: The other layouts all use the `DynamicList.prototype.focusCloseBtn` function declaration. That's cleaner and so please follow the same structure.

js/layout-javascript/simple-list-code.js
```diff
@@ -2428,6 +2435,11 @@ DynamicList.prototype.showDetails = function(id, listData) {
     });
   }
 
+  $(_this.$detailsContent).mousedown(function(event) {
```
tonytlwu, line 2438: This doesn't take into account that a user may use TAB to bring their focus to a link within the overlay, then press ENTER. That will also end up scrolling the user to the top.

tonytlwu (changes requested):

PR #687, opened 2022-04-14
OD-692 [Fix] The field type for the Simple List description has been changed to html by default

OleksandrLavrynenko:
@romanyosyfiv

OD-692 https://weboo.atlassian.net/browse/OD-692

## Description
**Problem**: A field type for the description of Simple List is set to be plain text by default but the data contains HTML 
**Solution**: The field type for the Simple List description has been changed to html by default

## Screenshots/screencasts
https://user-images.githubusercontent.com/86256215/163371227-e8e868d6-3dd6-40cf-93f9-6c17e713fe22.mp4

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


tonytlwu (changes requested):
You should be able to simple set https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/6e1ba47fcdfd74218b904654bef52a8123fa0bbe/js/default-configs/layouts-config.js#L348 to `html`

OleksandrLavrynenko:
> You should be able to simple set
> 
> https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/6e1ba47fcdfd74218b904654bef52a8123fa0bbe/js/default-configs/layouts-config.js#L348
> 
> to `html`
**@tonytlwu** I'm guessing I need to add an object to the "detail-fields"  https://github.com/Fliplet/fliplet-widget-dynamic-lists/blob/6e1ba47fcdfd74218b904654bef52a8123fa0bbe/js/default-configs/layouts-config.js#L362

image
what id should i add?

tonytlwu:
@OleksandrLavrynenko Any random 8 character ID with lower and upper case alphabets should be fine.

js/default-configs/layouts-config.js
```diff
@@ -365,6 +365,13 @@ window.flListLayoutConfig = {
         location: 'Image',
         type: 'image',
         column: 'Image'
+      },
+      {
+        id: 'RGosNfos',
+        interfaceName: 'Secondary text',
```
tonytlwu, line 371: @OleksandrLavrynenko I don't think `interfaceName` is needed. Can you confirm if or why it's necessary?

tonytlwu (changes requested):

js/default-configs/layouts-config.js
OleksandrLavrynenko: @tonytlwu You're right. This parameter is not needed

tonytlwu (approved):
@romanyosyfiv @inna-bieshulia This is ready for testing

Risk: Medium
Affected: Simple List LFD layouts
Verify: Detail views are configured with suitable defaults

PR #730, opened 2023-11-29
ID-3806 [Fix] Prevent legacy MixItUp classes from clashing with generic class names

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-3806


simonchapman1986 (approved):
Risk: low
Affected: dynamic lists
Verify: legacy mixitup classes don't clash with generic class names

PR #734, opened 2024-03-21
ID-4143 [fix] LFD data view settings reordering fields file

dwfliplet:
This PR changes the way the file fields are rendered in detailed views. Instead of being rendered at the end, they are rendered according to defined field order.

[ID-4143](https://weboo.atlassian.net/browse/ID-4143)

## Results

image

image

[ID-4143]: https://weboo.atlassian.net/browse/ID-4143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -2481,16 +2481,16 @@ DynamicList.prototype.addDetailViewData = function(entry, files) {
 
     if (obj.type === 'file') {
       if (files && Array.isArray(files)) {
+        var file = files.find(function(file) {
+          return file.id === obj.id;
+        });
 
-        files.filter(Boolean).forEach((file) => {
-          const fileAlreadyAdded = entry.entryDetails.some(({ id }) => id === file.id);
-
-          if (!fileAlreadyAdded) {
-            entry.entryDetails.push(file);
-          }
-        })
+        if (file) {
+          entry.entryDetails.push(file);
+        }
       }
-      return
+
+      return;
     }
 
     // Define label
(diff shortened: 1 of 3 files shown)


js/layout-javascript/news-feed-code.js
```diff
@@ -2480,6 +2480,16 @@ DynamicList.prototype.addDetailViewData = function(entry) {
     var content = '';
 
     if (obj.type === 'file') {
+      if (files && Array.isArray(files)) {
```
armine-fliplet, line 2483: @dwfliplet do we have a shared JS file where we can define a function for this code? We could then utilize this function across these three files to adhere to the DRY principle.

dwfliplet: @armine-fliplet Generally speaking, I agree that would be sufficient, but a big part of those files are duplicated. For example, the rest of the code is almost identical or identical in all three files. So there's no point in extracting a few lines when they're part of a bigger repeated sequence. To correctly manage duplications in those templates, I'd need to refactor those three files (>2k, >3k, and >3k lines), which I believe is out of the scope of this bugfix 😄.

But you're right, @armine-fliplet, that code needs refactoring in terms of duplications in order to decrease the amount of code to maintain. I think it's worth logging that in backlog, to plan maintenance in proper time. Wdyt @simonchapman1986?

simonchapman1986: So firstly, great to see your desire of keeping code dry @armine-fliplet - its certainly something we want and need.

However, I do think what @dwfliplet makes sense in this context - we shouldn't bring into scope a large refactor as this is a bugfix. If this was feature dev, i'd be more inclined to say - lets do the refactor. But we want to get this out the door quickly and efficiently.

I would say its worth adding a low priority item for the backlog noting this however, so we can indeed give it some time when we feel appropriate.

armine-fliplet (approved):

PR #790, opened 2026-04-12
PS-1808 [FIX] Reject negative string values in admin column check

tonytlwu:
## Summary
- **Security fix**: `userIsAdmin()` used JavaScript truthiness to check admin status, so any non-empty string (including "No", "N", "false", "0") granted admin access
- Added a blocklist of common negative strings that are rejected before the truthiness check
- Positive values ("X", "Yes", "Admin", "1", custom strings) continue to work — fully backward compatible
- Updated the interface tooltip to clarify that negative values won't grant access

## Test plan
- [x] Added `__tests__/user-is-admin.test.js` with 72 tests covering positive values, blocked negatives, empty/nil, arrays, whitespace trimming, and exact-match mode
- [x] All tests pass (`npx jest`)
- [ ] Manual verification in a Fliplet app with LFD: Admin="X" grants access, Admin="No" denies, Admin=blank denies

```
Deploy Check — #790 (fliplet-widget-dynamic-lists)
======================================
Ticket:    PS-1808 — Admin column in DS grants admin access when any value entered
Jira Status: [3rd] Waiting for Client Review
Risk:      LOW
QA:        Not signed off (Jira status is "Waiting for Client Review", not deploy-ready)
Reviews:   0/1 (none yet)
CI:        test: PASS | Codacy: ACTION_REQUIRED (non-blocking)
Mergeable: Clean

Blast Radius:
- Direct: 3 files changed (utils.js +15/-4, interface.html +1/-1, new test file)
- Consumers: 3 functions (recordIsEditable, recordIsDeletable, userCanAddRecord)
- Features: All LFD add/edit/delete permission checks across all 5 layouts
- Cross-service: No

### DIFF
--- a/__tests__/user-is-admin.test.js
+++ b/__tests__/user-is-admin.test.js
@@ -1,0 +1,168 @@
+/**
+ * Tests for PS-1808: userIsAdmin should not grant access for negative string values
+ *
+ * Verifies that values like "No", "N", "false", "0" do not grant admin access,
+ * while positive values like "X", "Yes", "Admin" continue to work.
+ */
+
+var fs = require('fs');
+var path = require('path');
+
+// Extract the functions from utils.js by evaluating them with minimal mocks
+var NativeUtils = {
+  get: function(object, prop) {
+    if (object === null || object === undefined) {
+      return undefined;
+    }
+
+    return object[prop];
+  },
+  isNil: function(value) {
+    return value === null || value === undefined;
+  }
+};
+
+// Extract userIsAdmin and isNegativeAdminValue by evaluating the relevant portion
+var utilsSource = fs.readFileSync(path.join(__dirname, '..', 'js', 'utils.js'), 'utf8');
+
+// Extract NEGATIVE_ADMIN_VALUES, isNegativeAdminValue, and userIsAdmin
+var negativeValuesMatch = utilsSource.match(/var NEGATIVE_ADMIN_VALUES = (\[.*?\]);/);
+var NEGATIVE_ADMIN_VALUES = eval(negativeValuesMatch[1]); // eslint-disable-line no-eval
+
+function isNegativeAdminValue(value) {
+  return typeof value === 'string'
+    && NEGATIVE_ADMIN_VALUES.indexOf(value.trim().toLowerCase()) > -1;
+}
+
+function userIsAdmin(config, userData) {
+  var adminValue = NativeUtils.get(userData, config.userAdminColumn);
+
+  if (NativeUtils.isNil(config.userAdminValue) || config.userAdminValue === '') {
+    if (Array.isArray(adminValue)) {
+      return !!adminValue.find(function(v) {
+        return v && !isNegativeAdminValue(v);
+      });
+    }
+
+    return !!adminValue && !isNegativeAdminValue(adminValue);
+  }
+
+  if (Array.isArray(adminValue)) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by Arpanexe — CHANGES_REQUESTED** [review]
## PR Review: PS-1808 [FIX] Reject negative string values in admin column check

### Summary

This PR fixes a security bug where `userIsAdmin()` relied on JavaScript truthiness to determine admin status. Any non-empty string — including `"No"`, `"N"`, `"false"`, `"0"` — would evaluate as truthy and grant admin access. The fix adds a blocklist of common negative strings (`NEGATIVE_ADMIN_VALUES`) that are rejected before the truthiness check. An updated tooltip in `interface.html` clarifies the behavior. A comprehensive test file with 72 tests is included.

### Critical Issues (Must Fix Before Merge)

**1. Test file uses `eval()` and re-implements source functions instead of testing the actual code**
`__tests__/user-is-admin.test.js:28` — The test extracts `NEGATIVE_ADMIN_VALUES` from the source file using `eval()`:
```js
var NEGATIVE_ADMIN_VALUES = eval(negativeValuesMatch[1]);
```
Then it manually re-implements `isNegativeAdminValue` and `userIsAdmin` (lines 30–48) rather than importing/requiring them from `utils.js`. This means:
- The tests don't actually test the real code — they test a copy
- If someone changes the logic in `utils.js` but forgets to update the test copy, tests will still pass
- The `eval()` is fragile and a code quality flag (likely what Codacy flagged)

**Recommendation:** Since `utils.js` registers via `Fliplet.Registry.set('dynamicListUtils', ...)`, the test should mock `Fliplet.Registry` and then `require('../js/utils.js')` to get the actual exported functions. This is the pattern that would properly test the real code path.

### Warnings (Should Address)

**1. Blocklist may be incomplete for international users**
The `NEGATIVE_ADMIN_VALUES` list covers English negative words. Fliplet apps are used internationally — values like `"nein"`, `"non"`, `"não"` could also cause the same original bug (user intends "no" but gets admin). Consider whether this is a known concern or document the limitation.

**2. `adminValue` of type `number` (non-zero) bypasses the blocklist**
If `adminValue` is the number `2` or `-1`, `isNegativeAdminValue` returns `false` (since `typeof value === 'string'` fails), and `!!adminValue` is `true`. This is probably fine for the intended use case, but worth noting that the blocklist only applies to strings.

**3. Codacy check is failing**
The CI shows `Codacy Static Code Analysis: fail`. This is likely due to the `eval()` usage in the test file. While the PR description says it's non-blocking, it should ideally be resolved.

### Suggestions (Nice to Have)

- The test file is well-structured with good coverage of edge cases (whitespace trimming, arrays, exact-match mode, nil values). Good work on the thoroughness.
- Consider adding `"nope"` to the blocklist — it's a casual but not uncommon English negative that could be entered by users.
- The tooltip update in `interface.html:480` is a good UX touch. Could also mention that leaving the field blank denies access (which is the intended pattern for non-admins).

### Verdict

**REQUEST_CHANGES** — The core logic change in `utils.js` is correct, minimal, and backward-compatible. The security fix is sound. However, the test file needs to be reworked to test the actual exported functions rather than a re-implemented copy. The `eval()` + manual function duplication pattern means these tests don't actually validate the production code, which undermines the value of the test suite. Fix the test approach, and this is ready to merge.

Arpanexe (approved):
LGTM

PR #792, opened 2026-04-25
PS-1885 [FIX] LFD bookmarks should work offline

tonytlwu:
## Summary

- Pass `offline: true` to the `LikeButton` instance behind each layout's bookmark control (simple-list, news-feed, small-card, agenda).
- This unblocks the "Please connect to the internet" popup that was firing every time a user tapped a bookmark while offline, so the toggle now reaches the data source connector which already handles offline queueing and reconnect replay.
- Like buttons are intentionally left as-is — they remain online-only because the like count is a shared online concept. Bookmarks are a per-user toggle that users (and the support docs) expect to keep working offline.

Fixes [PS-1885](https://weboo.atlassian.net/browse/PS-1885).

## Companion PR

This depends on the `offline` option being added to `LikeButton`:
- `fliplet-api` PR: https://github.com/Fliplet/fliplet-api/pull/7948

## Test plan

- [ ] In each layout (simple-list, news-feed, small-card, agenda): airplane mode → tap a bookmark → no popup; bookmark icon flips to filled state; bookmarked-only filter still works.
- [ ] Reconnect → confirm the bookmark/unbookmark write was synced server-side (check the bookmarks data source).
- [ ] Like button on a news-feed list still shows the "Please connect to the internet" popup when offline (regression check).
- [ ] Cold offline relaunch (with bundled DS): bookmark state from previous online session still renders.

[PS-1885]: https://weboo.atlassian.net/browse/PS-1885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/layout-javascript/agenda-code.js
+++ b/js/layout-javascript/agenda-code.js
@@ -2318,7 +2318,8 @@ DynamicList.prototype.setupBookmarkButton = function(options) {
           addType: 'prepend',
           getAllCounts: false,
           liked: record.bookmarked,
-          silent: record.bookmarkButton
+          silent: record.bookmarkButton,
+          offline: true
         });
 
         if (record.bookmarkButton) {
--- a/js/layout-javascript/news-feed-code.js
+++ b/js/layout-javascript/news-feed-code.js
@@ -2578,7 +2578,8 @@ DynamicList.prototype.setupBookmarkButton = function(options) {
           likedWrapper: '<div class="news-feed-bookmark-wrapper btn-bookmarked focus-outline" tabindex="0"></div>',
           addType: 'html',
           getAllCounts: false,
-          liked: record.bookmarked
+          liked: record.bookmarked,
+          offline: true
         });
 
         record.bookmarkButton = btn;
--- a/js/layout-javascript/simple-list-code.js
+++ b/js/layout-javascript/simple-list-code.js
@@ -2313,7 +2313,8 @@ DynamicList.prototype.setupBookmarkButton = function(options) {
           likedWrapper: '<div class="simple-list-bookmark-wrapper btn-bookmarked focus-outline" tabindex="0"></div>',
           addType: 'html',
           getAllCounts: false,
-          liked: record.bookmarked
+          liked: record.bookmarked,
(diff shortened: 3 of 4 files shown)


tonytlwu:
## Review

**Verdict:** approve.

### Categories applied

- **Scope correctness** — All four layouts that support bookmarks were touched (simple-list, news-feed, small-card, agenda). `small-h-card-code.js` has no bookmark support (verified with grep — no `setupBookmarkButton`, no `LikeButton`, no `userBookmarks` references), so it correctly stays unmodified.
- **Like vs Bookmark split** — The companion `setupLikeButton()` calls in `simple-list-code.js:2108` and `news-feed-code.js:2373` are intentionally left without `offline: true`. Like remains online-only, as described. This is the right call — a like count is a shared online concept, a bookmark is a per-user toggle.
- **Race conditions** — toggle-bookmark-twice-while-offline behaviour relies on the connector queue. Verified offline-replay path is correct in the api PR review.
- **Companion PR alignment** — depends on the `offline` opt being added to `LikeButton`. The api PR (Fliplet/fliplet-api#7948) initially patched only `0.1/like.js` but the widget loads `fliplet-like:0.2` per `widget.json:69`. That gap was caught during review and fixed in api commit `9f12574`. Both 0.1 and 0.2 now support the flag, so this widget change is correctly wired up.

### Informational (not blocking)

- **No regression test added.** The change is a one-line option propagation in each layout. The `__tests__/` directory has no offline-behaviour tests today and adding one needs an offline harness — out of scope for this PR. Recommend a manual smoke test on each layout per the PR test plan.
- **Bookmark identity drift** — pre-existing concern: bookmarks toggled offline before login attach to `device.uuid`, never migrate to the user's account on login. Not introduced by this PR; worth a follow-up ticket.

### Cross-model coverage

Single-model review (Codex CLI not available locally). The change is low-risk surface — four identical edits, all opt-in.

Arpanexe (approved):
LGTM

PR #794, opened 2026-06-10
PS-1879 [FIX] Detail view fields no longer repopulate after deletion

Komsomol:
## Summary
With "I want the detail view to auto update when new fields are added" enabled, the detail-view merge re-added **any** data source column missing from `detailViewOptions` — including columns the user had intentionally removed. Deleted fields therefore reappeared when the Data View Settings overlay was re-opened, and re-rendered in preview at runtime.

Fix tracks a `detailViewKnownColumns` snapshot (the DS column set as of the last save) and only auto-adds columns that are **not** in it. Intentionally-removed columns remain in the snapshot, so they are treated as "known" and never re-added; genuinely new columns (absent from both the saved options and the snapshot) still appear. Existing configs with no snapshot seed it from the current columns, so previously-deleted fields are not re-added.

Fixes [PS-1879](https://weboo.atlassian.net/browse/PS-1879)

## Acceptance criteria
- [x] Checkbox ON: deleting fields + Save & Close → fields stay gone on re-open — `js/interface.js` `knownColumns` gate
- [x] Checkbox ON: deleted fields do not re-render in preview — all 5 runtime layouts gate `extraColumns` on `knownColumns`
- [x] Genuinely new DS columns still auto-appear — new columns are in neither `savedColumns` nor `knownColumns`
- [x] Checkbox OFF: deletions persist (unchanged) — changes live inside the existing `if (detailViewAutoUpdate)` blocks
- [x] Existing apps without a snapshot don't re-add deleted fields — seed `|| dataSourceColumns`

## Test plan
- [ ] App **454728 / page 1936384** (QA repro), checkbox ON: delete Buttons/Order/Popular → Save & Close → reopen Data View tab → fields stay deleted; preview omits them
- [ ] Add a new column to the data source → confirm it auto-appears (feature preserved)
- [ ] Checkbox OFF → deletions persist (no regression)
- [ ] Spot-check across layouts (news-feed, simple-list, small-card, small-h-card, agenda)

## Verification (automated)
- Tests: 72 passed, 0 failed (`npx jest`)
- Lint: no new errors/warnings (2 parsing errors + `no-console` warnings are pre-existing on master)
- Changed: `js/interface.js` + 5 runtime layout files — all `detailViewAutoUpdate` merge sites gated

[PS-1879]: https://weboo.atlassian.net/browse/PS-1879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/detail-view-auto-update.test.js
+++ b/__tests__/detail-view-auto-update.test.js
@@ -1,0 +1,85 @@
+/**
+ * Tests for PS-1879: detail-view fields must not repopulate after deletion.
+ *
+ * When "auto update when new fields are added" is on, the merge logic only adds
+ * columns that are genuinely new to the data source. A column the user removed is
+ * recorded in the detailViewKnownColumns snapshot, so it is excluded from the
+ * re-add. These tests cover the two primitives that combine to produce that
+ * behaviour (NativeUtils.difference variadic + NativeUtils.coalesceArray) and the
+ * empty-snapshot edge case that would otherwise silently resurrect the bug.
+ */
+
+/* eslint-env jest */
+
+var path = require('path');
+
+// native-utils.js assigns to `window.NativeUtils`; expose `window` then require it.
+global.window = global.window || {};
+require(path.join(__dirname, '..', 'js', 'native-utils.js'));
+
+var NativeUtils = global.window.NativeUtils;
+
+describe('PS-1879 — NativeUtils.difference (variadic exclusion)', function() {
+  it('returns only columns absent from BOTH saved options and the known snapshot', function() {
+    var dsColumns = ['name', 'email', 'phone', 'newCol'];
+    var savedColumns = ['name', 'email'];
+    var knownColumns = ['name', 'email', 'phone']; // phone deleted by user: known, not saved
+
+    expect(NativeUtils.difference(dsColumns, savedColumns, knownColumns)).toEqual(['newCol']);
+  });
+
+  it('re-adds nothing when every data source column is already known', function() {
+    var dsColumns = ['a', 'b'];
+
+    expect(NativeUtils.difference(dsColumns, ['a'], ['a', 'b'])).toEqual([]);
+  });
+});
+
+describe('PS-1879 — NativeUtils.coalesceArray (empty-snapshot guard)', function() {
+  it('keeps a non-empty snapshot', function() {
+    expect(NativeUtils.coalesceArray(['a'], ['x', 'y'])).toEqual(['a']);
+  });
+
+  it('falls back when the snapshot is an empty array (truthy-empty footgun)', function() {
+    expect(NativeUtils.coalesceArray([], ['x', 'y'])).toEqual(['x', 'y']);
+  });
+
+  it('falls back when the snapshot is absent (existing apps with no snapshot)', function() {
+    expect(NativeUtils.coalesceArray(undefined, ['x', 'y'])).toEqual(['x', 'y']);
+    expect(NativeUtils.coalesceArray(null, ['x', 'y'])).toEqual(['x', 'y']);
+  });
+});
+
+describe('PS-1879 — column resolution end to end', function() {
+  // The deleted column ("phone") still exists in the data source but was removed
+  // from the detail view. It must NOT be re-added on any subsequent load/render.
+  function newColumnsToAdd(snapshot, dsColumns, savedColumns) {
+    var knownColumns = NativeUtils.coalesceArray(snapshot, dsColumns);
+
+    return NativeUtils.difference(dsColumns, savedColumns, knownColumns);
+  }
+
+  it('does not re-add a deleted column when a real snapshot exists', function() {
+    var snapshot = ['name', 'email', 'phone'];
+    var dsColumns = ['name', 'email', 'phone'];
+    var savedColumns = ['name', 'email']; // user deleted phone
+
+    expect(newColumnsToAdd(snapshot, dsColumns, savedColumns)).toEqual([]);
+  });
+
+  it('does not re-add a deleted column when the snapshot is empty (the Arpan hazard)', function() {
+    var snapshot = []; // persisted before columns loaded
+    var dsColumns = ['name', 'email', 'phone'];
+    var savedColumns = ['name', 'email']; // user deleted phone
+
+    expect(newColumnsToAdd(snapshot, dsColumns, savedColumns)).toEqual([]);
+  });
+
+  it('still surfaces a genuinely new data source column once a snapshot exists', function() {
+    var snapshot = ['name', 'email']; // columns known as of last save
+    var dsColumns = ['name', 'email', 'phone']; // phone added to the DS afterwards
+    var savedColumns = ['name', 'email'];
+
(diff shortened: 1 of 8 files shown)


Komsomol:
Ready for review — widget fix → `master`, CI green (test + Codacy pass).

**What changed:** gated the `detailViewAutoUpdate` merge on a new `detailViewKnownColumns` snapshot so intentionally-deleted detail-view fields are no longer re-added — in `interface.js` (edit overlay) and all 5 runtime layouts.

**Please check two judgment calls:**
1. Migration for existing apps: when there's no snapshot yet, it's seeded from current columns (`|| dataSourceColumns` / `|| _this.dataSourceColumns`). Effect — deleted fields stay gone immediately without a re-save, but auto-add of genuinely-new columns is dormant for that config until its first re-save in Studio. Acceptable trade-off?
2. `small-h-card-code.js` builds `savedColumns` from `.data` rather than `.column` (pre-existing quirk). Left untouched — the `knownColumns` guard fixes the deletion bug there regardless. Say if you'd rather fix it in this PR.

**QA (not automated here):** verify on app **454728 / page 1936384**, checkbox ON → delete fields → Save & Close → reopen + preview stay gone; then add a new DS column → confirms it still auto-appears.

@Arpanexe @galisufyan-327 — one of you for code review please.

galisufyan-327 (dismissed):
LGTM

Arpanexe:
## Review: PS-1879 — Detail view fields no longer repopulate after deletion

### Overview

With "auto update detail view" enabled, the merge logic re-added any data source column missing from `detailViewOptions`, so intentionally deleted fields came back. The fix persists a `detailViewKnownColumns` snapshot (DS columns as of last save) and only auto-adds columns absent from **both** the saved options and the snapshot. The gate is applied at all six merge sites: the Studio overlay (`js/interface.js`) and the five runtime layouts.

### What I verified (not just read)

- **`NativeUtils.difference(array, ...values)` is variadic** (`js/native-utils.js:454`) — it flattens all exclusion arrays into one Set, so the new three-argument call `difference(dataSourceColumns, savedColumns, knownColumns)` behaves exactly like lodash. This was the highest-risk assumption in the PR and it holds.
- **`dataSourceColumns` is in scope at the save site** — it's a module-scope `var` (`interface.js:12`), already used a few lines above the new code (`data.dataSourceColumns = dataSourceColumns` at line 2942).
- **The early `return` for known columns skips nothing else** — in the original loop, columns found in `detailViewOptions` already returned early; the only side effect was the push for missing columns, which is precisely what the gate is meant to suppress.
- The approach is consistent across all five layouts, including `small-h-card-code.js`, which maps `savedColumns` from the `'data'` property rather than `'column'` — a pre-existing inconsistency this PR correctly leaves untouched.

### Issues

**1. Empty-snapshot hazard silently resurrects the bug (worth fixing before merge).**
At the save site:

```js
data.detailViewKnownColumns = dataSourceColumns || _this.config.dataSourceColumns || _this.config.defaultColumns || [];
```

If a save happens before the data source columns have loaded (or the DS fetch failed), this persists `[]`. An empty array is **truthy**, so on the next load `_this.config.detailViewKnownColumns || dataSourceColumns` resolves to `[]`, no column is "known", and every missing column — including deleted ones — gets re-added. The original bug returns with no error and no obvious cause. Suggest treating an empty snapshot as absent at both read sites, e.g.:

```js
var knownColumns = (_this.config.detailViewKnownColumns && _this.config.detailViewKnownColumns.length)
  ? _this.config.detailViewKnownColumns
  : dataSourceColumns;
```

…or refusing to overwrite an existing non-empty snapshot with an empty one at save time. One guard in either place closes the hole.

**2. Behavioral tradeoff for existing apps should be stated more explicitly.**
For configs with no snapshot, the fallback seeds `knownColumns` from the *current* columns — which means the auto-update feature is effectively **paused** for every existing app (overlay and runtime) until the widget settings are re-saved once. Genuinely new DS columns added in that window won't auto-appear. The PR body presents the seed as protection against re-adding deleted fields (true), but the acceptance criterion "genuinely new DS columns still auto-appear" only holds *after* a snapshot exists. This is probably the right tradeoff — re-adding deleted fields is the bug being fixed, missing a new column is recoverable — but QA and the PR description should say it plainly.

### Minor

- **No new tests.** The diff touches 6 files, none of them tests; "72 passed" is the pre-existing suite. The merge gate in `interface.js` and especially the empty-snapshot edge case from issue 1 are cheap to unit-test and are exactly the kind of regression that won't be caught manually.
- **The same 4-line comment is copy-pasted five times** across the layout files. One sentence per site ("PS-1879: exclude columns the user removed; see interface.js for the snapshot logic") would carry the same information.
- Snapshot is written even when the checkbox is OFF — this is actually good (deletions survive a later toggle to ON), just noting it's intentional and correct.

### Verdict

**Approve with one requested change.** The design is sound — the snapshot correctly distinguishes "removed by user" from "new to the data source", the variadic `difference` call works, and all merge sites are covered. Fix the empty-array seed (issue 1) before merging, since it silently reintroduces the exact bug this PR fixes; a length-guard at the two read sites or the one write site is a two-line change.

Komsomol:
Thanks @Arpanexe — all addressed in cc098f0.

**Issue 1 (empty snapshot) — fixed.** You're right that `[]` is truthy and would silently resurrect the bug. Added `NativeUtils.coalesceArray(value, fallback)` (returns `value` only when it's a non-empty array, else `fallback`) and used it at all six read sites, so an empty/absent snapshot falls back to the current DS columns → deleted columns stay excluded. The write site can still persist `[]` but that's now harmless since every read coalesces.

**Minor: tests — added.** `__tests__/detail-view-auto-update.test.js` (8 tests): `difference()` variadic exclusion, `coalesceArray`, and the empty-snapshot regression specifically (`snapshot=[]` + deleted column → not re-added). Suite now 80 passing.

**Minor: duplicated comment — collapsed** to one line per layout file, pointing at `interface.js` for the rationale.

**Issue 2 (tradeoff) — documented.** Added a "Behaviour note for existing apps" section to the PR description: the auto-add feature is effectively paused for a config until its first re-save (which persists the snapshot); re-adding deleted fields is the bug, missing a brand-new column until re-save is recoverable. Flagged for QA too.

**`small-h-card` `.data` quirk** — left as-is per your agreement; the `knownColumns` guard fixes the deletion bug there regardless. Happy to file a separate ticket for the `.data` vs `.column` mapping if you think it's worth it.

Komsomol:
📌 **Behaviour note for existing apps** (couldn't edit the PR description — token issue — so recording it here for QA):

For a config with **no snapshot yet** (any app created before this fix), `knownColumns` falls back to the *current* data source columns:
- ✅ Previously-deleted fields are **not** re-added — immediately, no re-save needed.
- ⚠️ "Auto-add new fields" is effectively **paused for that config until its settings are saved once** (the save persists the snapshot). A brand-new DS column added in that window won't auto-appear until the next save.

Intended tradeoff: re-adding deleted fields is the reported bug; missing a new column until the next save is recoverable. So **AC "new columns still auto-appear" holds only after a snapshot exists** — QA should test that path *after* an initial save, not on a pre-existing config.

Arpanexe (approved):
LGTM

PR #795, opened 2026-08-06
DEV-1686 [FIX] Dynamic Lists no longer ships expiring timezone data (DST breaks 15 Mar 2027)

absolutehype:
## Summary

The widget bundles its own copy of moment-timezone (`vendor/`, listed in `widget.json` `build.assets`). Because it is baked into the build output it never passes the dependency resolver, and it loads **after** everything else on the screen — so whatever zone data it carries silently becomes the data every other script on that screen sees.

That copy was **0.5.33 / IANA 2021a**, whose DST table ends **November 2026**. From **15 March 2027** every DST-observing zone would render an hour early, with no error and nothing in the console. Future-dated content was already wrong today: a 2027 agenda rendered now was already an hour out.

Replaced with a custom **moment-timezone 0.6.3 / IANA 2026c** build spanning **2015–2100**.

Fixes [DEV-1686](https://weboo.atlassian.net/browse/DEV-1686)

## Why a custom build, and why the span runs to 2100

Upstream's full-data build fixes the bug but is 720KB. Measuring where the bytes actually live changed the shape of the fix:

| | raw | gzip |
|---|---|---|
| Old bundled build (0.5.33 / 2021a, ends 2026) | 42 KB | 11.5 KB |
| Upstream full data (1970–2499) | 720 KB | 36.5 KB |
| **This build (2015–2100)** | **78 KB** | **13.8 KB** |

Two findings drove it:

- **History is where the cost is.** Extending the *start* year back from 2020 to 2000 costs **+3,323 bytes** gzipped. Trimming it is what takes 720KB down to 78KB.
- **The future is nearly free.** Extending the *end* year from 2036 to 2100 — 64 extra years — costs **+831 bytes** gzipped, because once DST rules become periodic the transitions compress to almost nothing.

So capping the span short saves very little and is exactly what caused this bug. The span ends at 2100 deliberately. `filterLinkPack` also collapses 340 zones into 143 zones + 454 links once post-filter duplicates merge.

Net cost versus the (broken) build on master: **+2.3 KB gzipped**.

## Version is deliberately NOT bumped

`widget.json` stays at `1.4.0`. `fliplet-api/routes/v1/widgets.js:727` matches an existing widget by **exact** `package` + `version` (the major-version-only variant applies only to `type:function` widgets; this is `type:component`). Publishing `1.5.0` would create a **new widget row with a new `widgetId`**, existing widget instances would keep pointing at the old row, and the recompile step — which queries `where: { widgetId: <new id> }` — would find zero instances. The fix would reach new apps only.

For context, this widget has been `1.4.0` since January 2019 across ~750 PRs.

## Regenerating the vendor file

Not committed to avoid expanding scope — happy to add it under `tools/` if reviewers prefer. Requires `moment` and an extracted `moment-timezone` tarball:

```js
const moment = require('moment');
global.moment = moment;
require('moment-timezone/builds/moment-timezone-with-data.min.js');
require('moment-timezone/moment-timezone-utils.js');

const p = require('moment-timezone/data/packed/latest.json');
// filterLinkPack needs an UNPACKED bundle; npm ships only the packed one
const unpacked = {
  version: p.version,
  zones: p.zones.map((z) => moment.tz.unpack(z)),
  links: p.links.slice(),
  countries: p.countries.map((c) => {
    const parts = c.split('|');
    return { name: parts[0], zones: parts[1].split(' ') };
  })
};
const filtered = moment.tz.filterLinkPack(unpacked, 2015, 2100);
// then: header comment + builds/moment-timezone.min.js + moment.tz.load(filtered)
```

The file's own header comment records the library version, IANA version, span and source.

## Test plan

- [x] Reproduces then fixes the ticket's canary: `moment('2029-06-12T14:00:00Z').tz('America/New_York')` → was `09:00 -05:00`, now **`10:00 -04:00`**
- [x] DST horizon (America/New_York) moves from `2026-11-01` to `2100-11-07`
- [x] Verified under Fliplet's actual load pattern (global `moment`, no module/define): tz attaches, and both `js/utils.js` call sites work — `moment.tz.guess()` returns a zone name, `moment.tz(x, 'UTC')` resolves
- [x] Vendor file is a faithful `filterLinkPack` output of upstream `moment-timezone@0.6.3` packed data, not a hand-edited blob
- [ ] On a published screen containing a list, run the ticket's QA snippet — expect data `2026c` and `10:00 -04:00`
- [ ] Compare against a screen with no list in the same app (both should now agree)
- [ ] Exercise a date filter with **"use device timezone"** enabled plus a `now`/`today` modifier (`js/utils.js:1476`) — this is the path that would break silently if zone data were ever missing
- [ ] Confirm the publish reaches **all regions** (EU/US/CA) — the upload route restricts to the current region when an `organizationId` is set

## Verification (automated)

- Tests: **90 passed, 4 suites** (80 pre-existing + 10 new)
- Lint: new test file is **clean (0 problems)**; the repo's 119 existing problems are pre-existing on master, mostly in two other test files
- **Negative control**: swapping the old 2021a build back in fails **4 tests** — the horizon check, the header-comment check and two DST conversions. The guard is not vacuous.
- `moment@2.29.4` added as a devDependency (pinned to the platform's version) so the test can load moment-timezone; CI already runs `npm ci && npm test`

## Notes for reviewers

- The vendor file was **renamed** to `moment-timezone-with-data.min.js`. The old name claimed a 10-year range; `fliplet-widget-push-notifications` kept that stale name after swapping in full data, and the mismatch is a large part of what made this bug hard to trace. Referenced in exactly one place.
- **Related, not fixed here:** the widget still overwrites screen-wide `moment.tz` rather than sharing a single copy. Doing that properly needs a data-bearing `moment-timezone` registered in `fliplet-api`'s `config/common/assets.json` first — the entry there today is version `0.5.21` with **no zone data at all** (6,222 bytes), so simply deleting this vendored copy and declaring the dependency would leave `moment.tz` with no data and break device-timezone date filters silently. Worth a separate low-priority platform ticket.
- **Unrelated but worth flagging:** DEV-1686's description contains a live session auth token in plain text. It should be revoked and stripped.

[DEV-1686]: https://weboo.atlassian.net/browse/DEV-1686


galisufyan-327 (changes requested):
Code review (via `/fliplet-pr-review`)

The investigation behind this PR is first-rate — the root cause, the load-order analysis, the byte-cost measurements and the version-bump reasoning all hold up. I verified the two platform-side claims independently against `fliplet-api` and **both are accurate**:

- `routes/v1/widgets.js:727-733` matches on exact `package` + `version`; the major-version-only variant is gated on `type:function`, and `widget.json:6` tags this `type:component`. Keeping `1.4.0` is correct. The recompile step at `routes/v1/widgets.js:417` does query `where: { widgetId: res.flWidget.id }`.
- `config/common/assets.json:1274-1278` declares `moment-timezone` version `0.5.21` → `moment-timezone.min.js` (the no-data build). So **DEV-1686's premise is wrong** — "the platform copy is already correct, Dynamic Lists just needs to stop shipping its own" is not an available fix. The asset the ticket observed at load position 38 is another widget's vendored copy, not the platform asset. Worth correcting on the ticket.

Raising one **Critical** issue on the replacement data file, detailed inline.

---

## Critical — 139 of 597 zone names do not resolve in the new build

The payload does contain 143 zones + 454 links (matching the description), but `moment.tz.names()` returns only **458**. Loading the committed file under the browser pattern (global `moment`, no `module`/`define`) and comparing against the deleted build and upstream 0.6.3:

```
committed build : lib 0.6.3  data 2026c  names 458
deleted build   : lib 0.5.33 data 2021a  names 593
upstream 0.6.3  : lib 0.6.3  data 2026c  names 597
```

**137 names that work on master no longer resolve.** Converting `2029-06-12T14:00:00Z`:

```
Europe/Amsterdam   master: 15:00 +01:00   this PR: no data -> falls back to local time
America/Toronto    master: 09:00 -05:00   this PR: no data -> falls back to local time
Asia/Taipei        master: 22:00 +08:00   this PR: no data -> falls back to local time
America/Winnipeg   master: 08:00 -06:00   this PR: no data -> falls back to local time
Europe/Warsaw      master: 15:00 +01:00   this PR: no data -> falls back to local time
```

Also affected: `America/Regina`, `America/Montreal`, `America/Tijuana`, `Europe/Zurich`, `Europe/Prague`, `Europe/Stockholm`, `Europe/Copenhagen`, `Europe/Helsinki`, `Europe/Rome`, `Europe/Oslo`, `Asia/Riyadh`, `Asia/Kuwait`, `Asia/Ho_Chi_Minh`, `GMT`, `Etc/GMT` — +124 more.

### Root cause

`filterLinkPack` merges post-filter duplicate zones into links but carries `input.links` through unchanged, so a name can be both an alias and a base:

```
America/New_York|America/Toronto      <- derived by the duplicate merge
America/Toronto|America/Montreal      <- carried over from upstream
America/Toronto|Canada/Eastern
```

moment-timezone's `addLink` writes into a flat `links` map (last write wins) and `getZone` deliberately resolves **only one level** of indirection (the `caller !== getZone` guard). So `America/Toronto` resolves to `America/Montreal`, which is not a zone, and returns `null`. Same for chains like `Europe/Amsterdam -> Europe/Brussels -> Europe/Paris`. In the committed data: **113 links whose base is itself a link**, and **48 aliases that are also a base elsewhere**.

### Why this is worse than the bug being fixed

The original bug is a one-hour error after Nov 2026. This is *no data at all* for 139 zones, starting on merge — and because the file overwrites `moment.tz` screen-wide (the exact mechanism this PR identifies), it hits every component on the screen, not just Dynamic Lists. `js/utils.js:1381-1382` (`moment.tz.guess()` -> `setDefault`) degrades to browser-local time, which happens to be right when the target zone *is* the device zone, so the device-timezone filter path mostly survives; any explicit non-device zone is silently wrong.

### Fix — two options, both verified

**(a) Flatten the links after `filterLinkPack`** so every alias points directly at a surviving zone. Applied to the committed payload, this restores full coverage:

```js
const zoneNames = new Set(output.zones.map((z) => z.split('|')[0]));
const base = new Map();

output.links.forEach((l) => {
  const [b, a] = l.split('|');

  if (!base.has(a)) base.set(a, b);
});

function root(name, depth) {
  if (zoneNames.has(name) || depth > 10) return name;

  const b = base.get(name);

  return b ? root(b, depth + 1) : name;
}

output.links = [...new Set(output.links.map((l) => {
  const alias = l.split('|')[1];

  return root(alias, 0) + '|' + alias;
}))].filter((l) => {
  const [b, a] = l.split('|');

  return zoneNames.has(b) && !zoneNames.has(a);
}).sort();
```

Result: `names = 597`, and `America/Toronto -> 10:00 -04:00`, `Europe/Amsterdam -> 16:00 +02:00`, `Asia/Taipei -> 22:00 +08:00` — all correct.

**(b) Ship upstream's full-data build instead.** `fliplet-widget-push-notifications/vendor/moment-timezone-with-data-10-year-range.min.js` already *is* the upstream full-data build (0.6.0 / 2025b, **597 names**, DST horizon **2499**). There's already precedent in the codebase, it needs no custom tooling, and it removes this class of bug permanently. Cost is ~23KB gzipped over the custom build, against a `rendering.js` the ticket measures at 582KB. Given that a hand-rolled build just silently dropped a quarter of the world's timezones, the byte-optimisation looks like a poor trade.

### Related — the guard tests can't detect this

All four DST assertions in `__tests__/timezone-data.test.js:81-103` use zones that survive as **base** zones; `moment.tz.names()` is never asserted and no alias is exercised. I ran the suite on the branch: **90 passed, 4 suites** — matching your claim, and fully green despite 139 broken zones. The negative control is real, but it only proves the tests catch a stale *span*, not missing *zones*. Suggest adding:

```js
it('resolves every zone name it advertises', function() {
  var unresolved = moment.tz.names().filter(function(name) {
    return !moment.tz.zone(name);
  });

  expect(unresolved).toEqual([]);
  expect(moment.tz.names().length).toBeGreaterThanOrEqual(590);
});

it('resolves link zones, not just base zones', function() {
  expect(moment('2029-06-12T14:00:00Z').tz('America/Toronto').format('HH:mm Z'))
    .toBe('10:00 -04:00');
  expect(moment('2029-06-12T14:00:00Z').tz('Europe/Amsterdam').format('HH:mm Z'))
    .toBe('16:00 +02:00');
});
```

---

CI is green (Codacy pass, `test` pass) — which is precisely the problem noted above.

absolutehype:
Confirmed and fixed in 6089ffa. Thank you — this was a real defect and the diagnosis was exact.

## Reproduced independently

Your count of 139 is precise. Loading the committed file under the browser pattern:

```
committed build : names 458   (139 missing)
upstream 0.6.3  : names 597
```

All twelve zones you listed returned `null` and fell back to local time. Root cause is exactly as you described: `createLinks` seeds with `links = source.links.slice()` (upstream's 257 aliases verbatim) and then *appends* links derived from post-filter duplicate zones, never reconciling the two — so a name ends up both alias and base, and `getZone`'s one-level guard returns null for the chain.

## Fix

Took option **(a)**, flattening links after `filterLinkPack` so every alias points directly at a surviving zone, with a cycle guard and a hard throw if any alias can't be resolved. I went this way rather than **(b)** because the trimmed payload was a deliberate call after measuring the cost curve, and the failure was a generator bug rather than something inherent to trimming. **(b) remains entirely reasonable** if you'd still rather not carry custom tooling — say the word and I'll swap it.

I've also recorded the pitfall in the vendor file's header comment, since `filterLinkPack` alone looks like it works and the next person regenerating would otherwise walk straight back into it.

## Verified exhaustively, not by spot-check

Spot-checking three base zones is what let this through, so I compared **every** advertised zone against upstream 0.6.3 full data at 30 instants spanning 2015–2099:

```
FIXED vs upstream
  names advertised : 597 vs 597
  missing          : 0
  zones differing  : 0
  cells differing  : 0 / 17910
```

Size is essentially unchanged: 78,509 raw / **13,879 gzipped**.

## On the suggested tests — one subtlety

Adopted, with an adjustment. `expect(unresolved).toEqual([])` **passes on the broken build**, so it wouldn't have caught this on its own:

```
✓ resolves every name it advertises      <- passes on the BROKEN build
✕ advertises at least 590 zone names
✕ resolves link zones, not just base zones
✕ converts correctly through an alias
```

`moment.tz.names()` filters to `zones[i] || zones[links[i]]` — one level — so broken zones are *omitted from the list* rather than appearing in it and failing. The count floor and the explicit alias assertions are what do the work. I kept your check anyway (it guards a different failure mode) and noted the reasoning in the test file so it isn't "simplified" away later.

Four tests added, run against three builds:

| Build | Result |
|---|---|
| Fixed | **94 passed** |
| Broken (this PR's first commit) | **3 failed** |
| Original 2021a (master) | **5 failed** |

## On the ticket

Agreed, and already flagged — DEV-1686's premise doesn't hold, and the PR description covers why the "stop shipping your own" fix isn't available while `assets.json` resolves to the no-data build. I'll correct the ticket directly.

Your closing point stands and is the real lesson here: CI was green with a quarter of the world's timezones missing. The guard tests now cover coverage as well as span.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

galisufyan-327 (approved):
LGTM

